Issue #135, opened 2026-04-09
Daemon TCP connection breaks immediately on first event (macOS aarch64)


haixuanTao:
## Description

When running a Rust node with adora (both v0.2.1 tag and latest main `d59fa5d7`), the daemon TCP connection breaks immediately on the first event. The node never receives an `Input` event — instead, the very first `events.recv()` returns an `Error`:

```
Error("fatal event stream error: daemon channel broken\n\nCaused by:\n   0: failed to deserialize DaemonReply\n   1: io error: \n\nLocation:\n    apis/rust/node/src/daemon_connection/tcp.rs:61:14")
```

This happens with both `adora run dataflow.yml` and `adora up` + `adora start dataflow.yml`.

**Note:** This is NOT specific to `send_output()` or Arrow arrays — the connection breaks before any output is sent. The node never gets a chance to process data.

## Reproducer

Full reproducer code: https://github.com/dora-rs/dora-benchmark/tree/fix/adora-send-output-bug-repro/adora/rs-latency

Minimal node that reproduces the issue:

```rust
use adora_node_api::{self, AdoraNode, Event, adora_core::config::DataId, IntoArrow};

fn main() -> eyre::Result<()> {
    let latency = DataId::from("latency".to_owned());
    let (mut node, mut events) = AdoraNode::init_from_env()?;

    println!("waiting for first event");
    for i in 0..5 {
        match events.recv() {
            Some(Event::Input { metadata, .. }) => {
                println!("got tick i={i}");
                node.send_output(latency.clone(), metadata.parameters, 42u64.into_arrow())?;
            }
            Some(other) => {
                // Always lands here on i=0 with the daemon channel broken error
                println!("non-input event i={i}: {other:?}");
            }
            None => { println!("None at i={i}"); break; }
        }
    }
    Ok(())
}
```

### Dataflow

```yaml
nodes:
  - id: rust-node
    build: cargo build -p benchmark-example-node --release
    path: ./target/release/benchmark-example-node
    inputs:
      tick: adora/timer/millis/20
    outputs:
      - latency
      - throughput
  - id: rust-sink
    build: cargo build -p benchmark-example-sink --release
    path: ./target/release/benchmark-example-sink
    inputs:
      latency: rust-node/latency
      throughput: rust-node/throughput
```

### Steps to reproduce

```bash
git clone https://github.com/dora-rs/dora-benchmark.git
cd dora-benchmark && git checkout fix/adora-send-output-bug-repro
[...]

PR #137, opened 2026-04-09
ci: restore macOS example testing

### LINKED ISSUE #136: CI: restore macOS example/E2E testing (regression in dataflow execution on macOS aarch64)
## Problem

Dataflow execution is completely broken on macOS aarch64 (Apple Silicon). Running any example (including adora's own `rust-dataflow` example) immediately fails with:

```
fatal event stream error: daemon channel broken
  failed to deserialize DaemonReply
  apis/rust/node/src/daemon_connection/tcp.rs:61:14
```

This was not caught because macOS example testing was dropped during the CI rewrite.

## CI history

The old CI (commit `167ce121` and earlier) ran examples on all three platforms:

```yaml
examples:
    platform: [ubuntu-22.04, macos-latest, windows-latest]
    - name: "Rust Dataflow example"
      run: cargo run --example rust-dataflow
```

At commit `6a6cd0a4`, all CI workflows were disabled. When CI was rebuilt (`16e3df95` → `a281ac54`), the examples job was replaced with E2E tests that only run on Ubuntu. The macOS `cargo test` job passes because it only runs unit tests with examples excluded — it never actually executes a dataflow.

## Impact

- adora does not work at all on macOS aarch64
- No example or E2E test catches this because none run on macOS
- Related: #135

## Fix

Restore cross-platform example testing in CI. PR incoming.

haixuanTao:
## Summary
- Restore cross-platform example testing that was dropped during the CI rewrite
- Add an `examples` job running on both `ubuntu-latest` and `macos-latest`
- Runs `rust-dataflow` and `multiple-daemons` examples end-to-end

## Context

The old CI ran examples on all platforms (ubuntu, macos, windows). When CI was rebuilt, example testing was replaced with E2E tests that only run on Ubuntu. This left macOS without any end-to-end dataflow testing.

As a result, a daemon TCP deserialization regression on macOS aarch64 went undetected — dataflow execution is completely broken on Apple Silicon (see #135).

This PR will **intentionally fail on macOS** until #135 is fixed, which is the point — it makes the regression visible in CI.

Fixes #136

## Test plan
- [ ] CI runs the `examples` job on both Ubuntu and macOS
- [ ] Ubuntu examples should pass
- [ ] macOS examples will fail (expected — demonstrates the bug in #135)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,39 @@ jobs:
           cargo test --test fault-tolerance-e2e -- --test-threads=1
         timeout-minutes: 15
 
+  examples:
+    name: Examples (${{ matrix.platform }})
+    runs-on: ${{ matrix.platform }}
+    needs: test
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
+      fail-fast: false
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Build CLI
+        run: cargo build -p adora-cli --release
+      - name: Install CLI (Linux)
+        if: runner.os == 'Linux'
+        run: cp target/release/adora /usr/local/bin/adora
+      - name: Install CLI (macOS)
+        if: runner.os == 'macOS'
+        run: cp target/release/adora /usr/local/bin/adora
+      - name: Build examples
+        timeout-minutes: 30
+        run: cargo build --examples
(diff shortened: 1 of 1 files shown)


haixuanTao:
## Related issues for full CI restoration

This PR is the first step. The full list of CI gaps from the rewrite is tracked in:

- #136 — Restore macOS example/E2E testing (this PR)
- #138 — Cross-platform example testing (Rust, C, C++)
- #139 — CLI integration tests (templates, Python, dynamic dataflows)
- #140 — Cross-compilation checks
- #141 — Benchmark example + MSRV check
- #142 — ROS2 bridge testing

heyong4725:
Superseded by #167, which covers all jobs from this PR plus the full upstream dora CI suite (examples, CLI, cross-check, ROS2, MSRV, license). Can be closed.

PR #158, opened 2026-04-09
fix(scheduler): preserve service/action correlations on drop_oldest

### LINKED ISSUE #145: `drop_oldest` queue policy silently drops service responses and action results
The default queue policy (`drop_oldest`) can silently discard service responses carrying
`request_id` or action results carrying `goal_id`/`goal_status`. The scheduler drops messages
purely by position with no awareness of communication pattern metadata, breaking the
request-response correlation contract.

## Problem

At `apis/rust/node/src/event_stream/scheduler.rs:164-176`, when a queue is full:

```rust
queue.pop_front();  // Drops oldest — no check if it carries request_id or goal_status
This means:

A service client waiting for a response with a specific request_id may wait forever because the response was dropped from its input queue
An action client may never receive the final result message with goal_status, leaving the goal state machine in limbo
No error, warning, or notification is generated — the drop is silent from the pattern's perspective

heyong4725:
## Summary

Fixes the silent correlation drop reported in #145. The default `drop_oldest` queue policy could discard service responses (`request_id`) and action results (`goal_id` / `goal_status`) before the client ever had a chance to receive them, leaving request/response state machines in limbo with no error or warning.

## Root cause

`Scheduler::add_event` at `apis/rust/node/src/event_stream/scheduler.rs:174` called `queue.pop_front()` unconditionally whenever the queue hit its cap. The scheduler had no awareness of pattern metadata — a service response carrying `request_id` got the same treatment as a discardable sensor reading.

## Fix — preserve-when-possible eviction

A new `select_eviction` helper walks the queue front-to-back and prefers evicting non-correlated events. Three cases:

1. **Queue has a non-correlated event** → drop the oldest such event, keep all correlations (including the head).
2. **Queue is entirely correlated but the incoming is not** → drop the incoming event. This breaks strict drop_oldest semantics for one message but is the only way to preserve existing correlations without an OOM-risk policy change.
3. **Queue is entirely correlated AND incoming is also correlated** → unavoidable. Evict the oldest and emit `tracing::error!` with the dropped `request_id` / `goal_id` / `goal_status` so operators can trace which correlation was broken and respond (increase `queue_size` or switch to `queue_policy: backpressure`).

The `dropped` counter still increments in every case, so `drain_drop_counts` keeps working for observability.

## What counts as "correlated"

Metadata keys from `libraries/message/src/metadata.rs`:

- `REQUEST_ID` (`"request_id"`) — service request/response
- `GOAL_ID` (`"goal_id"`) — action goal
- `GOAL_STATUS` (`"goal_status"`) — action result

These are the well-known keys the service and action helpers already use; no new protocol surface is added.

## Tests

Five new regression tests in `scheduler::tests`:

| Test | Scenario |
|---|---|
| `drop_oldest_preserves_correlated_when_non_correlated_present` | head is correlated, middle is non-correlated → middle gets dropped |
| `drop_oldest_drops_middle_non_correlated_to_save_front_correlated` | two correlations sandwich one non-correlated → the sandwich filling is dropped |
| `drop_oldest_drops_incoming_if_queue_is_fully_correlated_and_incoming_is_not` | queue is 100% correlated → incoming non-correlated is dropped instead |
| `drop_oldest_drops_front_loudly_when_both_queue_and_incoming_are_correlated` | unavoidable-drop path, verifies the oldest is evicted and new one admitted |
| `drop_oldest_goal_id_is_also_preserved` | same preservation via `goal_id` (not just `request_id`) |

Existing `drop_oldest_tracks_drop_count` and `backpressure_policy_*` tests still pass — no behavior change for the non-correlated default path.

## Test plan

- [x] `cargo test -p adora-node-api --lib event_stream::scheduler::` — 11/11 pass (6 existing + 5 new)
- [x] `cargo test -p adora-node-api` — full suite clean

Fixes #145

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -4,9 +4,85 @@ use adora_message::{
     config::{DEFAULT_QUEUE_SIZE, QueuePolicy},
     daemon_to_node::NodeEvent,
     id::DataId,
+    metadata::{GOAL_ID, GOAL_STATUS, REQUEST_ID, get_string_param},
 };
 
 use super::thread::EventItem;
+
+/// Returns `true` if the event carries request/response or action correlation
+/// metadata (`request_id`, `goal_id`, or `goal_status`).
+///
+/// These keys bind a message to a specific service request or action goal.
+/// Silently dropping such a message breaks the correlation contract — the
+/// client waits forever for a response or result that never arrives
+/// (dora-rs/adora#145).
+fn is_correlated(event: &EventItem) -> bool {
+    let EventItem::NodeEvent {
+        event: NodeEvent::Input { metadata, .. },
+        ..
+    } = event
+    else {
+        return false;
+    };
+    let params = &metadata.parameters;
+    params.contains_key(REQUEST_ID)
+        || params.contains_key(GOAL_ID)
+        || params.contains_key(GOAL_STATUS)
+}
+
+/// Outcome of `select_eviction`.
+enum Eviction {
+    /// Remove event at this index from the queue and push the incoming event.
+    RemoveAt(usize),
+    /// The queue is entirely correlated and the incoming event is not —
+    /// drop the incoming event instead of breaking a correlation.
+    DropIncoming,
+    /// The queue is entirely correlated and the incoming event is also
+    /// correlated — drop the oldest (front) event with a loud error log.
+    DropFrontLoud,
+}
+
+/// Choose which event to drop when the queue is at capacity.
+///
+/// Prefers sacrificing non-correlated events so that service responses and
+/// action results survive. See `is_correlated` for the metadata keys that
+/// mark an event as part of a pattern.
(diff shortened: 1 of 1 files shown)


heyong4725:
LGTM.

I reviewed the diff and did not find a blocking issue.

What I checked:
- `apis/rust/node/src/event_stream/scheduler.rs` now preserves service/action-correlated inputs by preferring eviction of non-correlated messages when a `drop_oldest` queue is full.
- When the queue is entirely correlated, the code falls back to dropping the oldest correlated message with an explicit error log, which is the right failure mode for an unavoidable overflow.
- The added tests cover the important cases: preserving `request_id`, preserving `goal_id`, dropping incoming non-correlated messages when the queue is fully correlated, and the unavoidable correlated-drop case.

Validation:
- `cargo test -p adora-node-api scheduler` passed on the PR 158 branch.

One non-blocking note: this intentionally relaxes strict same-input FIFO eviction when correlated and non-correlated messages are mixed, because it may evict a middle non-correlated message instead of the front. That looks like the correct tradeoff for preserving service/action correlation.

If CI is green, I would merge PR #158.

PR #161, opened 2026-04-09
fix(daemon,node-api): guarantee NodeRestarted delivery and tracker handling

### LINKED ISSUE #148: Node restart orphans in-flight service requests and action goals
When a service-server or action-server node restarts via the fault tolerance system, pending
`request_id` correlations and active `goal_id` state machines are silently abandoned. Clients
waiting for responses or results will block indefinitely.

## Problem

At `binaries/daemon/src/lib.rs:3326-3355`, the restart path:
1. Sends `NodeRestarted` to downstream nodes — but this event can be **dropped** if the
   channel is full (line 3346: `Ok(false) => { /* event dropped (channel full) */ }`)
2. Does **NOT** clean up pending `request_id`s — no cancellation or timeout notification
   is sent to service clients
3. Does **NOT** set `goal_status` to `"aborted"` or `"canceled"` — action clients never
   receive a terminal goal status
4. The restarted server has no knowledge of pre-crash requests/goals

## Impact

- Service clients wait forever for responses that will never come
- Action clients' goal state machines hang in a non-terminal state
- Even if `NodeRestarted` is delivered, the client has no way to know which specific
  `request_id`s or `goal_id`s were lost
- The `InputTracker` at `apis/rust/node/src/event_stream/input_tracker.rs:60-77` does not
  handle `NodeRestarted` events (returns `false`), so cached state is not cleared

heyong4725:
## Summary

First of two PRs addressing #148 (Option C from the design discussion on the issue).

Scope here is the **observability + lifecycle plumbing** layer:

1. Daemon never silently drops `NodeRestarted` anymore — falls back to a bounded-await send on try_send failure.
2. `InputTracker` now handles `NodeRestarted` when given an opt-in source map.

The **pattern-aware client helpers** (`recv_service_response(request_id, timeout)` and `recv_action_result(goal_id, timeout)`) and the docs update on `docs/patterns.md` will follow in a separate PR, since they add new public API surface and are orthogonal to this fix.

## (1) Daemon: guaranteed NodeRestarted delivery

`binaries/daemon/src/lib.rs` at the `SpawnedNodeResult` / restart path previously absorbed `Ok(false)` from `send_with_timestamp` with an empty comment `/* event dropped (channel full) */`. The internal `send_with_timestamp` *does* already log a `CRITICAL: control event dropped` line in that case, but the caller then continued as if nothing happened — no context about which dataflow, restarted node, or receiver, and no delivery retry. Downstream clients silently stayed unaware of the restart, leaving service/action clients holding orphaned `request_id` / `goal_id` correlations forever.

New behavior on the restart notification path:

1. **Fast path unchanged.** Still calls `send_with_timestamp` first (benefits from the existing `CONTROL_EVENT_HEADROOM` reservation).
2. **On `Ok(false)`**: emit a structured error with dataflow id, restarted node, receiver id — then fall back to a backpressure-aware `channel.send(msg).await` wrapped in a 500ms `tokio::time::timeout` so the daemon main loop can't stall on a stuck receiver.
3. **Three fallback outcomes**: delivered (bump pending), channel closed (warn), timeout (CRITICAL error with full context).

The fallback only engages when the control-headroom-protected try_send already failed, which is rare in practice but must not be silently absorbed when it happens.

## (2) InputTracker: handle NodeRestarted with optional source map

`apis/rust/node/src/event_stream/input_tracker.rs` previously returned `false` for `NodeRestarted`. The honest reason: the tracker had no mapping from `NodeId → DataIds this node feeds`, so it couldn't know which inputs to mark as recovered.

New opt-in API:

```rust
let mut source_map = HashMap::new();
source_map.insert(DataId::from("sensor"), NodeId::from("camera"));
let tracker = InputTracker::with_source_map(source_map);
```

With the map populated, `process_event(&Event::NodeRestarted { id })` iterates entries and transitions any `Closed` inputs sourced from the restarted node back to `Healthy`. Cached `last_value` is preserved so nodes can still fall back to stale data between restart and the first post-restart message.

**Backward compat**: `InputTracker::new()` still works. Without a source map, `NodeRestarted` is acknowledged (returns `true` — so callers inspecting the boolean don't treat the lifecycle signal as noise) but state is not mutated. Existing users get the same behavior until they opt in.

## Regression tests

Four new tests in `input_tracker::tests`:

| Test | Scenario |
|---|---|
| `node_restarted_without_source_map_is_acknowledged_but_noop` | closed input stays closed, but `process_event` returns `true` |
| `node_restarted_with_source_map_recovers_matching_closed_inputs` | only inputs sourced from the restarted node recover; other inputs untouched |
| `node_restarted_preserves_last_value_cache` | stale cached data survives restart so graceful-degradation paths still work |
| `node_restarted_leaves_healthy_inputs_alone` | already-Healthy inputs are not churned |

All 4 existing tracker tests still pass.

Daemon-side change is verified by `cargo test -p adora-daemon` (no new test — the restart path is deep inside `handle_event` and would require a full daemon fixture; the logic is straightforward and the observability improvement is itself the value-add).

## Test plan

- [x] `cargo clippy -p adora-node-api -p adora-daemon -- -D warnings`
- [x] `cargo test -p adora-node-api` — all tests pass (4 existing tracker + 4 new)
- [x] `cargo test -p adora-daemon` — all tests pass

## Follow-up (tracked in #148)

- Add `AdoraNode::recv_service_response(request_id, timeout)` that surfaces `NodeRestarted` as `ServiceError::ServerRestarted` and `timeout` as `ServiceError::Timeout`
- Add `AdoraNode::recv_action_result(goal_id, timeout)` equivalent
- Update `docs/patterns.md` with the fault-tolerance contract for correlated patterns and the remaining gap: the daemon does *not* synthesize per-correlation cancellations; clients must use the new helpers (or their own timeout)

Partially fixes #148

### DIFF
--- a/apis/rust/node/src/event_stream/input_tracker.rs
+++ b/apis/rust/node/src/event_stream/input_tracker.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use adora_arrow_convert::ArrowData;
-use adora_core::config::DataId;
+use adora_core::config::{DataId, NodeId};
 
 use super::event::Event;
 
@@ -35,6 +35,12 @@ use super::event::Event;
 pub struct InputTracker {
     states: HashMap<DataId, InputState>,
     cache: HashMap<DataId, ArrowData>,
+    /// Optional input → source node map. When provided, `NodeRestarted`
+    /// events transition any `Closed` inputs sourced from the restarted
+    /// node back to `Healthy`. Without it the tracker has no way to know
+    /// which inputs a given upstream node feeds, so `NodeRestarted`
+    /// leaves state unchanged (dora-rs/adora#148).
+    source_map: HashMap<DataId, NodeId>,
 }
 
 /// Health state of a tracked input.
@@ -48,15 +54,39 @@ pub enum InputState {
 
 impl InputTracker {
     /// Create a new, empty tracker.
+    ///
+    /// Without a source map, `NodeRestarted` events are acknowledged
+    /// (return `true` from `process_event`) but do not mutate state,
+    /// because the tracker cannot determine which inputs a given node
+    /// feeds. If you use fault-tolerance restart policies, prefer
+    /// [`InputTracker::with_source_map`] so closed inputs recover
+    /// automatically when the upstream comes back up.
     pub fn new() -> Self {
         Self {
             states: HashMap::new(),
             cache: HashMap::new(),
+            source_map: HashMap::new(),
+        }
+    }
+
+    /// Create a tracker that knows which input is fed by which upstream
+    /// node. With this map populated, `NodeRestarted { id }` events
+    /// transition any `Closed` inputs sourced from `id` back to `Healthy`.
+    ///
+    /// The map can be built from the dataflow's `input_config` at
+    /// initialization time — for each user input `(data_id, source_node)`
+    /// insert the pair here.
+    pub fn with_source_map(source_map: HashMap<DataId, NodeId>) -> Self {
+        Self {
+            states: HashMap::new(),
+            cache: HashMap::new(),
+            source_map,
         }
     }
 
     /// Update tracker state from an event. Returns `true` if the event
-    /// was relevant to input tracking (Input, InputClosed, or InputRecovered).
+    /// was relevant to input tracking (`Input`, `InputClosed`,
+    /// `InputRecovered`, or `NodeRestarted`).
     pub fn process_event(&mut self, event: &Event) -> bool {
         match event {
             Event::Input { id, data, .. } => {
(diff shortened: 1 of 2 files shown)


heyong4725:
I found one blocking issue.

`binaries/daemon/src/lib.rs` still allows `NodeRestarted` to be dropped after the new fallback path times out.

What changed is an improvement over the old behavior:
- on `try_send` failure, the daemon now falls back to `timeout(Duration::from_millis(500), channel.send(msg)).await`
- that makes the failure louder and gives the event another chance to get through

But it does **not** guarantee delivery, because the timeout branch still just logs and returns:
- `binaries/daemon/src/lib.rs:3371` starts the bounded await
- `binaries/daemon/src/lib.rs:3389` logs the timeout and drops the event

So a slow or wedged receiver can still miss `NodeRestarted`, which is the exact failure mode the PR title says is fixed. If the intended contract is only “best effort with better observability,” this is fine, but then the PR should be reframed. If the intended contract is truly guaranteed delivery, this is still incomplete.

The `InputTracker` half looks good, and my local validation passed:
- `cargo test -p adora-node-api input_tracker`
- `cargo check -p adora-daemon`

I would not merge this as-is under the current title/claim.

heyong4725:
Re-checked on the latest PR branch state. The blocking issue I called out is fixed.

The daemon no longer uses the 500ms timeout fallback that could still drop `NodeRestarted`. It now falls back to an unconditional `channel.send(msg).await`, so the event is no longer lost on a slow receiver.

I re-ran the relevant checks on the updated branch:
- `cargo test -p adora-node-api input_tracker`
- `cargo check -p adora-daemon`

Both passed.

From my re-check, PR #161 is now mergeable.

PR #162, opened 2026-04-09
feat(node-api): add pattern-aware recv helpers with timeout + restart detection

### LINKED ISSUE #148: Node restart orphans in-flight service requests and action goals

heyong4725:
## Summary

Second of two PRs addressing #148 (Option C from the design discussion).

- #161 handled the **daemon observability + tracker lifecycle** side.
- This PR adds the **client-side pattern-aware helpers** that let service/action clients wait for a correlated reply without hanging forever when the server crashes.

Together the two PRs close #148.

No conflicts with #161 — it touches `input_tracker.rs` and `daemon/src/lib.rs`; this PR touches `event_stream/mod.rs`, `error.rs`, `lib.rs`, and `docs/patterns.md`. Either can merge first.

## New public API

```rust
pub use adora_node_api::PatternError;

// Service client:
let rid = node.send_service_request("request".into(), params, data)?;
match events
    .recv_service_response(&rid, &server_id, Duration::from_secs(5))
    .await
{
    Ok(Event::Input { data, .. }) => handle_response(data),
    Err(PatternError::Timeout) => fallback_path(),
    Err(PatternError::ServerRestarted(_)) => retry_with_new_instance(),
    Err(e) => return Err(e.into()),
    _ => unreachable!(),
}

// Action client:
events
    .recv_action_result(&goal_id, &server_id, Duration::from_secs(30))
    .await
```

`PatternError` variants: `Timeout`, `ServerRestarted(String)`, `StreamEnded`, `StreamError(String)`.

## Semantics

Both helpers drive `recv_async()` in a loop with a deadline. For each event:

| Event | Outcome |
|---|---|
| Matches caller's predicate | Return `Ok(event)` |
| `NodeRestarted { id } if id == expected_server` | Re-queue + `Err(ServerRestarted)` |
| `NodeRestarted` from *other* nodes | Buffer and keep waiting |
| `Stop(_)` | Re-queue + `Err(StreamEnded)` |
| `Error(_)` | `Err(StreamError)` |
| Deadline elapsed | `Err(Timeout)` |
| Anything else (other inputs, params, etc.) | Buffer and keep waiting |

Non-matching events are stashed in a new `pending_passthrough: VecDeque<Event>` field on `EventStream`. `recv_async` / `recv` drain this buffer first, so **the caller's main event loop never loses intermediate events** that arrived during a helper wait.

The action helper is stricter: it only matches on `goal_id` **and** a terminal `goal_status` (`succeeded` | `aborted` | `canceled`). Intermediate feedback (matching goal_id without terminal status) is passed through to the main loop so clients can observe progress.

## Testability refactor

Extracted the per-event decision into a pure `classify_correlation_event` free function returning a `CorrelationOutcome` enum. The async `wait_for_correlation` loop just drives `recv_async` and dispatches on the outcome. This lets the per-event logic be unit-tested without any `EventStream` plumbing.

## Tests

12 new unit tests in `event_stream::tests`:

**Service path (8):**
- `classify_matching_request_id_returns_match`
- `classify_different_request_id_is_passthrough`
- `classify_expected_server_restart_returns_server_restarted`
- `classify_unrelated_node_restart_is_passthrough`
- `classify_stop_returns_stream_ended`
- `classify_error_returns_stream_error`
- `classify_unrelated_input_is_passthrough`
- `classify_param_update_is_passthrough` — runtime params must survive helper waits

**Action path (4):**
- `classify_action_result_terminal_succeeded_matches`
- `classify_action_result_terminal_aborted_matches`
- `classify_action_feedback_without_terminal_status_is_passthrough`
- `classify_action_result_for_different_goal_is_passthrough`

All 18 tests in `event_stream::tests` pass (6 existing conversion + 12 new).

## Docs

`docs/patterns.md` updated in three places:

1. **Service section** — new "Waiting for a response with timeout + fault tolerance" subsection with full `recv_service_response` usage.
2. **Action section** — new "Waiting for a terminal result with timeout + fault tolerance" subsection with full `recv_action_result` usage.
3. **Important details** — new "Fault tolerance for correlated patterns" subsection documenting the limitation (the daemon does *not* synthesise per-correlation cancellations) and the recommended workaround (use the helpers), with a manual-handling snippet for callers who don't want the helpers.

## Test plan

- [x] `cargo test -p adora-node-api` — all tests pass (28 in event_stream module, including 12 new)

Fixes #148 (together with #161)

### DIFF
--- a/apis/rust/node/src/error.rs
+++ b/apis/rust/node/src/error.rs
@@ -22,3 +22,30 @@ pub enum NodeError {
 
 /// Convenience alias used by all public node API functions.
 pub type NodeResult<T> = std::result::Result<T, NodeError>;
+
+/// Errors returned by the pattern-aware receive helpers on
+/// [`EventStream`](crate::EventStream).
+///
+/// These cover the recoverable failure modes a service client or action
+/// client needs to react to when waiting for a correlated reply. See
+/// [`EventStream::recv_service_response`](crate::EventStream::recv_service_response)
+/// and
+/// [`EventStream::recv_action_result`](crate::EventStream::recv_action_result).
+#[derive(Debug, Error)]
+pub enum PatternError {
+    /// The correlated reply did not arrive before the deadline elapsed.
+    #[error("pattern reply timed out")]
+    Timeout,
+    /// The expected server node restarted, so the in-flight correlation
+    /// is orphaned. Clients should retry the request against the new
+    /// instance (dora-rs/adora#148).
+    #[error("server `{0}` restarted while waiting for reply")]
+    ServerRestarted(String),
+    /// The event stream ended (e.g. dataflow stopping) before the
+    /// correlated reply arrived.
+    #[error("event stream ended before pattern reply arrived")]
+    StreamEnded,
+    /// The event stream surfaced an error while waiting.
+    #[error("pattern wait failed: {0}")]
+    StreamError(String),
+}
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -22,7 +22,7 @@ use scheduler::{NON_INPUT_EVENT, Scheduler};
 
 use self::thread::{EventItem, EventStreamThreadHandle};
 use crate::{
-    DaemonCommunicationWrapper,
+    DaemonCommunicationWrapper, PatternError,
     daemon_connection::{DaemonChannel, node_integration_testing::convert_output_to_json},
     event_stream::data_conversion::{MappedInputData, RawData, SharedMemoryData},
 };
@@ -77,6 +77,12 @@ pub struct EventStream {
     /// Expected input types from YAML descriptor (for first-message validation).
     /// Each input is checked once; after the first message, the entry is removed.
     input_type_checks: HashMap<DataId, arrow_schema::DataType>,
+    /// Events that were consumed by a pattern-aware helper
+    /// (`recv_service_response`, `recv_action_result`) while searching
+    /// for a correlation match. Drained first on the next `recv()` so
+    /// the caller's main event loop never loses intermediate events
+    /// (dora-rs/adora#148).
+    pending_passthrough: std::collections::VecDeque<Event>,
 }
 
 impl EventStream {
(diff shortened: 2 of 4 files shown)


heyong4725:
Retrospective review: I found two behavioral regressions in the merged change.

1. `EventStream`s `Stream` implementation does not drain `pending_passthrough`, so callers using `StreamExt::next()` can permanently lose events buffered by `recv_service_response` / `recv_action_result`.

- `recv_async()` explicitly drains `pending_passthrough` first:
  `apis/rust/node/src/event_stream/mod.rs:512`
- but `impl Stream for EventStream` reads directly from `self.receiver` and bypasses that buffer:
  `apis/rust/node/src/event_stream/mod.rs:1019`

That means the PRs contract that the callers main event loop never loses intermediate events is only true for `recv` / `recv_async` / `try_recv` / `drain`, not for `StreamExt::next()`.

2. `EventStream::is_empty()` ignores `pending_passthrough`, so it can incorrectly report empty even when buffered passthrough events are still waiting to be drained.

- current implementation only checks `self.scheduler` and `self.receiver`:
  `apis/rust/node/src/event_stream/mod.rs:565`

That creates inconsistent public API behavior after a helper buffers unrelated events.

Validation I ran on the merged PR commit:
- `cargo test -p adora-node-api event_stream` passed

The existing tests cover the classification logic, but not the `pending_passthrough` integration path where these regressions sit. This should be fixed in a follow-up PR.

PR #164, opened 2026-04-09
fix(node-api): skip type checks on polymorphic pattern messages

### LINKED ISSUE #150: Type system ignores pattern polymorphism
## Summary

The type validation system assumes one Arrow type per output/input, but Service, Action,
and Streaming patterns legitimately send different schemas on the same output distinguished
by metadata. Enabling `ADORA_RUNTIME_TYPE_CHECK=error` can reject valid messages.

## Problem

**Runtime check** at `apis/rust/node/src/node/mod.rs:807-825`:
- Compares actual Arrow type against single declared `output_types` entry
- A service server sending different response schemas for different request types will
  have all but one type rejected

**First-message validation** at `apis/rust/node/src/event_stream/mod.rs:526-544`:
- Checks type on the FIRST message per input, then removes the check
- If first message is Action feedback (schema A) and later message is Action result
  (schema B), only A is validated

**Build-time check** at `libraries/core/src/descriptor/validate.rs:978-1007`:
- `check_metadata_annotations()` only validates presence of metadata KEYS, not that
  different metadata VALUES may require different schemas

**Pattern metadata** at `libraries/core/src/types.rs:380-387`:
- `pattern_metadata_keys()` returns key names but this info is never connected to type
  validation

## Impact

- `ADORA_RUNTIME_TYPE_CHECK=error` is unusable with polymorphic service/action servers
- First-message validation gives false confidence (only checks one variant)
- Build-time validation misses schema conflicts between pattern message variants

heyong4725:
## Summary

Fixes #150. Adora's type validation assumes one Arrow type per output/input, but service, action, and streaming patterns legitimately multiplex multiple Arrow schemas through a single channel. Both the `ADORA_RUNTIME_TYPE_CHECK` send-side gate and the first-message receive-side check rejected or spuriously warned on these polymorphic variants, making `ADORA_RUNTIME_TYPE_CHECK=error` unusable with service/action servers.

## Fix

Conservative exemption: when a message carries any of the pattern correlation keys (`request_id`, `goal_id`, `goal_status`), skip the type check. Non-pattern messages on the same output are still validated normally.

### Send side — `apis/rust/node/src/node/mod.rs`

New module-private helper:

```rust
pub(crate) fn carries_pattern_correlation(params: &MetadataParameters) -> bool {
    params.contains_key(adora_message::metadata::REQUEST_ID)
        || params.contains_key(adora_message::metadata::GOAL_ID)
        || params.contains_key(adora_message::metadata::GOAL_STATUS)
}
```

Called in `send_output` before the runtime type check. If it returns `true`, the check is skipped.

### Receive side — `apis/rust/node/src/event_stream/mod.rs`

The first-message type check previously used `remove(id)` which consumed the entry even on pattern messages, meaning the check fired exactly once on whatever variant happened to arrive first and never validated subsequent messages. Changed to peek with `get(id).cloned()` and only `remove(id)` when the first message is NOT a pattern message. The check stays armed across polymorphic variants and a later non-pattern message still validates.

## Docs

`docs/types.md` — "Runtime Type Checking" section updated with a new scope bullet:

> **Pattern messages are skipped.** When a message carries `request_id`, `goal_id`, or `goal_status` in its metadata parameters, runtime type checking is bypassed for that message. Service, action, and streaming patterns legitimately multiplex multiple Arrow schemas through a single output — a service server may reply with different response shapes for different request types — so a single declared Arrow type cannot cover all variants. Non-pattern messages on the same output are still validated normally.

Plus a note on the matching receive-side exemption.

## Tests

5 new unit tests in `node::tests`:

| Test | Scenario |
|---|---|
| `carries_pattern_correlation_detects_request_id` | service request/response |
| `carries_pattern_correlation_detects_goal_id` | action goal/feedback |
| `carries_pattern_correlation_detects_goal_status` | action terminal result |
| `carries_pattern_correlation_empty_is_not_a_pattern` | no keys → full validation |
| `carries_pattern_correlation_ignores_non_pattern_keys` | custom metadata keys don't suppress checks |

All 48 `adora-node-api` lib tests pass.

## Out of scope

The build-time `check_metadata_annotations` validator in `libraries/core/src/descriptor/validate.rs` could also warn when an output has `pattern: service-server` together with `output_types`, but the warning would be noisy for legitimate single-type uses and the user couldn't easily silence it. Leaving that for a later iteration once we have a way to declare polymorphism explicitly in YAML.

## Test plan

- [x] `cargo test -p adora-node-api --lib` — 48/48 pass

Fixes #150

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -525,21 +525,42 @@ impl EventStream {
 
         // First-message type validation: check once per input, then remove.
         // Zero cost after first message per input.
+        //
+        // Skip the check when the message carries pattern metadata
+        // (`request_id`, `goal_id`, or `goal_status`) — the input is
+        // polymorphic by pattern design and a single declared type
+        // cannot cover all variants. The check stays armed so a later
+        // non-pattern message can still validate (dora-rs/adora#150).
         if let Some(Event::Input {
-            ref id, ref data, ..
+            ref id,
+            ref metadata,
+            ref data,
         }) = event
-            && let Some(expected) = self.input_type_checks.remove(id)
+            && let Some(expected) = self.input_type_checks.get(id).cloned()
         {
-            let actual = data.data_type();
-            // Skip check for Null type (timer ticks, empty payloads)
-            // to avoid spurious warnings on annotated timer inputs.
-            if *actual != arrow_schema::DataType::Null && *actual != expected {
-                tracing::warn!(
-                    input = %id,
-                    expected = ?expected,
-                    actual = ?actual,
-                    "input type mismatch on first message"
-                );
+            let is_pattern_message = metadata
+                .parameters
+                .contains_key(adora_message::metadata::REQUEST_ID)
+                || metadata
+                    .parameters
+                    .contains_key(adora_message::metadata::GOAL_ID)
+                || metadata
+                    .parameters
+                    .contains_key(adora_message::metadata::GOAL_STATUS);
+            if !is_pattern_message {
+                // Consume the check and validate.
+                self.input_type_checks.remove(id);
+                let actual = data.data_type();
+                // Skip check for Null type (timer ticks, empty payloads)
+                // to avoid spurious warnings on annotated timer inputs.
+                if *actual != arrow_schema::DataType::Null && *actual != expected {
+                    tracing::warn!(
+                        input = %id,
+                        expected = ?expected,
+                        actual = ?actual,
+                        "input type mismatch on first message"
+                    );
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


heyong4725:
Retrospective review: I found one behavioral regression in the merged change.

The fix updates the `recv_async()` path, but not the direct `Stream` path, so polymorphic pattern messages can still spuriously consume the first-message type check when callers use `StreamExt::next()`.

What I checked:
- the new guarded logic is present in `apis/rust/node/src/event_stream/mod.rs:534`
- but `impl Stream for EventStream` still uses the old unconditional `remove(id)` logic in `apis/rust/node/src/event_stream/mod.rs:847`

That means the PRs intended behavior is only true for `recv` / `recv_async`, not for the `StreamExt::next()` path that the docs also support. If a pattern message arrives first on `StreamExt::next()`, the check is still consumed there and a later non-pattern message on the same input will never validate.

I did not see regression tests covering the `Stream` path variant of this behavior.

This should be fixed in a follow-up PR by making `Stream::poll_next` mirror the `recv_async()` pattern-message exemption logic and by adding regression tests for `StreamExt::next()`.

Issue #149, opened 2026-04-09
Circuit breaker (input_timeout) fires during normal service pattern idle periods


phil-opp:
## Summary

The `input_timeout` circuit breaker closes inputs based purely on time since last message.
For Service patterns where the server only sends responses when requests arrive, natural
idle periods on the response input trigger false-positive timeouts.

## Problem

At `binaries/daemon/src/lib.rs:1729-1761`:

```rust
if deadline.last_received.elapsed() > deadline.timeout {
    // Closes input — no awareness that this is a service response channel
    break_input(dataflow, node_id, input_id, &clock);
}
A service client's response input may legitimately receive no messages for extended
periods (no requests sent = no responses expected). The circuit breaker interprets this
silence as a failure and fires InputClosed.

Impact
False InputClosed events on service response inputs
InputTracker marks the input as Closed even though the server is healthy
If AllInputsClosed fires as a result, the client node may shut down unnecessarily

PR #178, opened 2026-04-10
feat(daemon): implement AddNode — spawn nodes into running dataflows

heyong4725:
Fixes #131. The AddNode handler was a stub since Sprint 10 that only logged a warning. Now implements the full spawn integration so `adora node add --from-yaml <file>` actually works.

## What changed

**Daemon** (`binaries/daemon/src/lib.rs`): Replaces the stub with a complete spawn path mirroring the per-node logic in `spawn_dataflow`:

1. Register inputs: `open_inputs`, `mappings` (User), `timers` (Timer), `log_subscribers` (Logs), `input_deadlines` (unarmed per #149)
2. Mark node in `pending_nodes` / `dynamic_nodes`
3. Create `Spawner`, call `spawn_node` -> prepare -> spawn -> `RunningNode`
4. Insert into `dataflow.running_nodes`

**Coordinator** (`binaries/coordinator/src/lib.rs`): Fixed the FIXME — replaced hardcoded `uv: false` with `dataflow.uv` so Python nodes added dynamically respect the original `--uv` flag.

## Test plan

- [x] `cargo clippy -p adora-daemon -p adora-coordinator -- -D warnings`
- [x] `cargo test -p adora-daemon --lib` — 27/27 pass
- [x] `cargo test -p adora-coordinator --lib` — 27/27 pass

Fixes #131

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1335,8 +1335,7 @@ async fn start_inner(
                                                                     DaemonCoordinatorEvent::AddNode {
                                                                         dataflow_id,
                                                                         node: resolved_node,
-                                                                        // FIXME: thread `uv` from ControlRequest::AddNode when available
-                                                                        uv: false,
+                                                                        uv: dataflow.uv,
                                                                     },
                                                                 timestamp: clock.new_timestamp(),
                                                             })?;
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1304,6 +1304,13 @@ async fn start_inner(
                                             node.id
                                         ))
                                     } else {
+                                        // Keep a clone of the original Node so
+                                        // we can push it into the stored
+                                        // descriptor after a successful spawn
+                                        // (so `adora info` reflects the new
+                                        // node).
+                                        let original_node = node.clone();
+
                                         // Resolve the Node into a ResolvedNode via a
                                         // temporary single-node descriptor.
                                         let tmp_desc = adora_message::descriptor::Descriptor {
@@ -1334,7 +1341,7 @@ async fn start_inner(
                                                                 inner:
                                                                     DaemonCoordinatorEvent::AddNode {
                                                                         dataflow_id,
-                                                                        node: resolved_node,
+                                                                        node: resolved_node.clone(),
                                                                         uv: dataflow.uv,
                                                                     },
                                                                 timestamp: clock.new_timestamp(),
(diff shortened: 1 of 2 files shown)


heyong4725:
I found two blocking issues in the AddNode implementation.

1. `binaries/daemon/src/lib.rs:1589` leaves the dataflow permanently inconsistent when dynamic node spawn fails. The handler mutates shared runtime state up front by inserting `open_inputs`, `mappings`/`timers`/`log_subscribers`, and `pending_nodes`/`dynamic_nodes`, then if `spawn_node`, build, or `prepared.spawn()` fails it only logs the error and returns. There is no rollback. For non-dynamic nodes, that means `pending_nodes` can stay set forever and prevent the dataflow from ever reaching completion; for both kinds, stale routing/deadline/log-subscriber state is left behind for a node that never actually exists.

2. `binaries/coordinator/src/lib.rs:1347` and `binaries/daemon/src/lib.rs:1603` never update the stored dataflow descriptor with the newly added node, so the control plane remains stale after a successful `AddNode`. The coordinator still serves `ControlRequest::Info` from `dataflow.descriptor.clone()` (`binaries/coordinator/src/lib.rs:941`), and the daemon still consults `dataflow.descriptor.nodes` for node metadata in the registration path (`binaries/daemon/src/lib.rs:2971`). As written, `adora node add` can spawn a working process, but `adora info` will not show the new node, and daemon behaviors that rely on descriptor lookup for that node remain inconsistent.

Validation:
- `cargo check -p adora-daemon -p adora-coordinator` passed on the PR 178 branch in a temporary worktree.

heyong4725:
Re-review on the updated PR 178 branch: one blocker remains.

The failed-spawn rollback issue is fixed. State mutations are now deferred until after spawn succeeds in `binaries/daemon/src/lib.rs:1543`, so I no longer see the stale `pending_nodes` / routing-state problem from my first review.

One blocking issue still remains:

`binaries/daemon/src/lib.rs:1533` still leaves the daemon's own descriptor state stale after a successful `AddNode`. The new follow-up commit fixes this on the coordinator side by pushing into `dataflow.descriptor.nodes` and `dataflow.nodes` there, but the daemon never mirrors that update. The daemon still consults `dataflow.descriptor.nodes` during node registration in `binaries/daemon/src/lib.rs:2971`, so dynamically added nodes are still missing from that lookup on the daemon that actually spawned them. This means the “descriptor stays consistent after add” blocker is only half-fixed.

Validation:
- `cargo check -p adora-daemon -p adora-coordinator` passed.

heyong4725:
Re-review on the latest PR 178 branch: one blocking consistency bug still remains.

`binaries/coordinator/src/lib.rs:1404` and `binaries/daemon/src/lib.rs:1716` still do not remove dynamically added nodes from stored descriptor/runtime metadata on `RemoveNode`. The new follow-up commits now push the node into coordinator `dataflow.descriptor.nodes` / `dataflow.nodes` and daemon `dataflow.descriptor.nodes` on `AddNode`, but the `RemoveNode` path only tears down `node_to_daemon` and runtime channels/process state. There is no inverse cleanup for those descriptor collections.

That means an add-then-remove cycle leaves stale nodes behind in `adora info`, and daemon descriptor-based lookups will continue to “see” a removed node. This is a new consistency bug now that PR 178 started mutating those descriptor stores.

Recommended fix:
- In coordinator `RemoveNode`, after successful daemon dispatch, also remove the node from `dataflow.descriptor.nodes` and `dataflow.nodes` alongside `node_to_daemon.remove(&node_id)`.
- In daemon `RemoveNode`, also remove the node from `dataflow.descriptor.nodes` alongside the existing runtime-state cleanup.
- Keep AddNode/RemoveNode symmetric: every collection mutated on add should have matching cleanup on remove.

Validation:
- `cargo check -p adora-daemon -p adora-coordinator` passed.

PR #179, opened 2026-04-10
feat(node-api): zero-copy zenoh receive path

### LINKED ISSUE #132: zenoh shared memory incomplete
There is an unconditional copy of the received data:

https://github.com/dora-rs/adora/blob/0cc06fce38f1bf25a7bb3d39af1c56f947d021c9/apis/rust/node/src/event_stream/mod.rs#L374-L401

So zero-copy messaging is not possible this way.

heyong4725:
Fixes #132. The zenoh subscriber path unconditionally copied received payloads into an `AVec` via `payload.to_bytes()`, making zero-copy messaging impossible even when Zenoh SHM was enabled.

## Fix

New `EventItem::ZenohInput` variant holds the raw zenoh `ZBytes` payload without copying. The Arrow conversion uses `Buffer::from_custom_allocation` to wrap the zenoh buffer directly:

- **SHM payloads** (`Cow::Borrowed`): true zero-copy — the Arrow buffer points directly into the shared memory region
- **Network payloads** (`Cow::Owned`): the owned `Vec` is wrapped at zero cost (no copy)

The `ZBytes` is kept alive as the `Arc` allocation owner via a `ZBytesAllocation` newtype that implements arrow's `Allocation` trait (`Send + Sync + RefUnwindSafe`).

## Changes

| File | Change |
|---|---|
| `event_stream/thread.rs` | New `EventItem::ZenohInput { id, metadata, payload: ZBytes }` |
| `event_stream/mod.rs` | Zenoh subscriber sends `ZenohInput`; new `zenoh_payload_to_arrow_array()`; `ZBytesAllocation` newtype |
| `event_stream/scheduler.rs` | `is_correlated()` and `add_event()` handle `ZenohInput` for flush/correlation |

No wire protocol changes — `ZenohInput` is `pub(crate)` on `EventItem`, which only exists inside the node process.

## Test plan

- [x] `cargo test -p adora-node-api --lib` — 75/75 pass
- [x] Smoke tests (`smoke_rust_dataflow*`) — 3/3 pass end-to-end

Fixes #132

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -1028,43 +1028,43 @@ fn zenoh_payload_to_arrow_array(
     // allocation owner. The pointer remains valid because the ZBytes
     // inside the Arc keeps the mapping alive.
     //
-    // For non-SHM payloads, to_bytes() returns Cow::Owned(Vec<u8>).
-    // We extract the pointer from the Vec, then move the ZBytes into
-    // the Arc owner. The Cow::Owned's Vec is NOT dropped because we
-    // use ManuallyDrop — its memory is owned by the ZBytes internally
-    // and will be freed when the Arc drops.
-    let cow = payload.to_bytes();
-    let (ptr, len) = match &cow {
-        std::borrow::Cow::Borrowed(slice) => (slice.as_ptr(), slice.len()),
-        std::borrow::Cow::Owned(vec) => (vec.as_ptr(), vec.len()),
-    };
-    // SAFETY: `ptr` points into memory owned by `payload` (either SHM
-    // mapping or internal Vec). We move `payload` into the Arc below
-    // so the allocation stays alive for the lifetime of the Buffer.
-    // The raw pointer doesn't carry the Cow borrow — it's just a number.
-    // We must drop `cow` before moving `payload`.
-    // Wrap the ZBytes in a newtype that satisfies arrow's Allocation
-    // trait (Send + Sync). ZBytes is Send but its Sync impl depends on
-    // zenoh internals. Since we only access the buffer from the thread
-    // that holds the Arrow array (no concurrent reads from multiple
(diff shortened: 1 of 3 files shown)


heyong4725:
I found one blocking issue in the new zenoh zero-copy receive path.

`apis/rust/node/src/event_stream/mod.rs:1036` creates a dangling Arrow buffer for any zenoh payload where `ZBytes::to_bytes()` returns `Cow::Owned`. The code takes `vec.as_ptr()` from that temporary owned buffer, then immediately `drop(cow)` before building `Buffer::from_custom_allocation` with `Arc::new(ZBytesAllocation(payload))` (`apis/rust/node/src/event_stream/mod.rs:1061`). That is only sound if the owned `Vec`'s allocation is actually backed by `payload`, but the API contract of `to_bytes()` allows it to materialize a fresh contiguous buffer for non-contiguous payloads. In that case the `Vec` is freed at `drop(cow)`, while Arrow keeps a pointer into freed memory. This turns the new path into use-after-free / UB for fragmented or copied zenoh buffers.

The SHM `Cow::Borrowed` case looks directionally correct, but the `Cow::Owned` branch needs different ownership handling.

Recommended fix:
- branch explicitly on `payload.to_bytes()`
- for `Cow::Borrowed(slice)`, keep the current `Buffer::from_custom_allocation(..., Arc<ZBytesAllocation(payload)>)` approach, because the pointer is into memory owned by `payload`
- for `Cow::Owned(vec)`, make the Arrow buffer own that `Vec<u8>` directly instead of `payload`, for example by using `Buffer::from_vec(vec)` or by using `from_custom_allocation` with the `Vec<u8>` itself as the allocation owner
- do not take a pointer out of a temporary owned `Cow` and then drop that owner before Arrow takes over

Validation:
- `cargo check -p adora-node-api` passed
- `cargo test -p adora-node-api event_stream -- --nocapture` passed

PR #183, opened 2026-04-10
fix(message): allow '/' in DataId for operator-namespaced outputs

### LINKED ISSUE #168: Multiple Daemons example: DataId contains invalid '/' character
## Problem

The `multiple-daemons` example fails with:

```
invalid DataId 'rust-operator/status': identifier contains invalid character '/' -- only [a-zA-Z0-9_.-] are allowed
```

at `libraries/message/src/id.rs:155`.

## Location

`examples/multiple-daemons/run.rs:244` — constructs a DataId with `/` which the ID validator rejects.

## How found

Restored CI job `examples` (PR #167) runs `cargo run --example multiple-daemons` on ubuntu + macos.

## Fix

Rename the DataId to use `-` or `_` instead of `/`, or adjust the operator output naming to not embed the operator path separator in the DataId.

heyong4725:
Fixes #168. Runtime-node operator outputs use `<operator-id>/<output-name>` (e.g. `rust-operator/status`). The DataId validator rejected `/` because it shared the same validator with NodeId.

## Fix

Split `validate_id` into two:
- `validate_node_id`: rejects `/` (unchanged — NodeIds can't have `/` since it's the node/output separator)
- `validate_data_id`: allows `/` for operator-namespaced outputs

## Verification

`adora validate examples/multiple-daemons/dataflow.yml` now passes (was crashing with `invalid DataId 'rust-operator/status'`).

## Tests

- `data_id_allows_slash_for_operators` — `rust-operator/status`, `op/output` both valid
- `data_id_rejects_other_specials` — spaces, semicolons still rejected
- All 10 id tests pass

Fixes #168

### DIFF
--- a/libraries/message/src/id.rs
+++ b/libraries/message/src/id.rs
@@ -3,8 +3,11 @@ use std::{borrow::Borrow, convert::Infallible, str::FromStr};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-/// Validate that an identifier contains only safe characters: `[a-zA-Z0-9_.-]`.
-fn validate_id(id: &str) -> Result<(), InvalidId> {
+/// Validate that a node identifier contains only safe characters: `[a-zA-Z0-9_.-]`.
+///
+/// NodeIds must NOT contain `/` because `/` is the separator between
+/// `<node_id>/<output_id>` in input mapping syntax.
+fn validate_node_id(id: &str) -> Result<(), InvalidId> {
     if id.is_empty() {
         return Err(InvalidId("identifier must not be empty".into()));
     }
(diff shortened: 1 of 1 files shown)


heyong4725:
I found one blocking issue in the current fix.

`libraries/message/src/id.rs` now allows slash-containing `DataId`s, which fixes the original `rust-operator/status` case from issue #168. However, it also accepts malformed values such as `"op/"`, `"/out"`, or `"a//b"`.

That becomes unsafe in the downstream runtime-node validation path:
- `libraries/core/src/descriptor/validate.rs:379` does `let (operator_id, output) = output.split_once(/) .unwrap_or_default();`
- `libraries/core/src/descriptor/validate.rs:381` then does `DataId::from(output.to_owned())`

For `"op/"`, that turns into `DataId::from("")`, which panics instead of producing a normal validation error. Before this PR, slash-containing `DataId`s were rejected up front, so this panic path was unreachable.

So I think the PR should still be blocked until one of these is done:
- reject malformed slash forms in `validate_data_id` (for example empty path segments / leading slash / trailing slash / double slash), or
- stop using panicking `DataId::from(...)` in the runtime descriptor validation path and return a proper error instead

I did verify that the original issue is addressed in the normal case: `cargo run --example multiple-daemons` got past the original `invalid DataId rust-operator/status` failure locally.

PR #196, opened 2026-04-12
fix(state): enforce strict runtime param delivery for running nodes

Bhanudahiyaa:
## Summary
  This PR makes runtime parameter mutation delivery strict for **running** targets.

  For `SetParam` / `DeleteParam`, success now requires daemon-confirmed runtime delivery to the target running node. Non-delivery paths now return errors instead of silently succeeding.

  ## Problem / Impact
  Previously, param mutations could report success even when runtime delivery did not happen (e.g. node not connected or channel full). That created control-plane/data-plane inconsistency:

  - CLI/API saw success
  - runtime node did not receive/apply the mutation
  - coordinator/store state suggested mutation succeeded

  This is a correctness issue across CLI → coordinator → daemon → node boundaries.

  ## What Changed

  ### 1) Daemon: strict delivery semantics for running nodes
  In `binaries/daemon/src/lib.rs`:

  - `SetParam` and `DeleteParam` now return daemon errors when delivery is not possible:
    - node not connected
    - node channel full
    - node channel closed
  - Added strict helpers:
    - `deliver_param_update_strict(...)`
    - `deliver_param_delete_strict(...)`

  ### 2) Coordinator: strict reply validation for running-target forwarding
  In `binaries/coordinator/src/lib.rs`:

  - Added reply validators:
    - `ensure_set_param_forward_applied(...)`
    - `ensure_delete_param_forward_applied(...)`
  - Running-target mutation path now requires:
    - `DaemonCoordinatorReply::SetParamResult(Ok(()))` for `SetParam`
    - `DaemonCoordinatorReply::DeleteParamResult(Ok(()))` for `DeleteParam`
  - Daemon rejection / unexpected reply now returns `ControlRequestReply::Error(...)` to caller.

  ### 3) E2E stability under strict startup timing
  In `tests/ws-cli-e2e.rs`:

  - Added bounded retry helper `set_param_with_retry(...)` for transient startup windows where node channels are not yet connected.
  - Updated param running-dataflow E2E tests to use retry helper.

  ## Why This Is High-Leverage
  This is a focused contract fix across system boundaries that improves correctness without new subsystem complexity:

  - aligns API success semantics with runtime application semantics
  - removes silent non-delivery success cases
  - improves reliability/debuggability in distributed state mutation flows

  ## Files Changed
  - `binaries/coordinator/src/lib.rs`
  - `binaries/daemon/src/lib.rs`
  - `tests/ws-cli-e2e.rs`

  ## Tests Added / Updated

  ### New targeted tests
  - Daemon:
    - strict update fails when node not connected
    - strict delete fails when channel full
  - Coordinator:
    - rejects daemon SetParam error reply
    - rejects unexpected reply variant for delete-forward validation

  ### Existing tests rerun
  - Full daemon suite
  - Full coordinator suite
  - Param-focused ws-cli E2E (`e2e_param_*`)

  ## Validation
  Ran following checks (all pass):

  - `cargo fmt --all -- --check`
  - `cargo clippy -p dora-daemon -p dora-coordinator -- -D warnings`
  - `cargo test -p dora-daemon -- --nocapture`
  - `cargo test -p dora-coordinator -- --nocapture`
  - `cargo test --test ws-cli-e2e e2e_param_ -- --nocapture --test-threads=1`

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1167,72 +1167,60 @@ async fn start_inner(
                             key,
                             value,
                         } => {
-                            let reply = match resolve_param_target(
-                                &running_dataflows,
-                                store.as_ref(),
-                                &dataflow_id,
-                                &node_id,
-                            ) {
-                                Ok(target) => match serde_json::to_vec(&value) {
-                                    Ok(bytes) => {
-                                        match store.put_node_param(
-                                            &dataflow_id,
-                                            &node_id,
-                                            &key,
-                                            &bytes,
-                                        ) {
-                                            Ok(()) => {
-                                                if let ParamTarget::Running { daemon_id } = target {
-                                                    if let Some(df) =
-                                                        running_dataflows.get_mut(&dataflow_id)
-                                                    {
-                                                        df.append_state_log(
-                                                            StateCatchUpOperation::SetParam {
-                                                                node_id: node_id.clone(),
-                                                                key: key.clone(),
-                                                                value: value.clone(),
-                                                            },
-                                                        );
+                            let reply: eyre::Result<ControlRequestReply> = async {
+                                let target = resolve_param_target(
+                                    &running_dataflows,
+                                    store.as_ref(),
+                                    &dataflow_id,
+                                    &node_id,
+                                )?;
+                                let bytes = serde_json::to_vec(&value)
+                                    .map_err(|e| eyre!("failed to serialize param value: {e}"))?;
+                                store.put_node_param(&dataflow_id, &node_id, &key, &bytes)?;
 
-                                                        if let Ok(msg) =
-                                                            serde_json::to_vec(&Timestamped {
-                                                                inner: DaemonCoordinatorEvent::SetParam {
-                                                                    dataflow_id,
-                                                                    node_id: node_id.clone(),
-                                                                    key: key.clone(),
-                                                                    value: value.clone(),
-                                                                },
-                                                                timestamp: clock.new_timestamp(),
-                                                            })
-                                                            && let Some(conn) =
-                                                                daemon_connections.get_mut(&daemon_id)
-                                                                && let Err(e) =
-                                                                    conn.send_and_receive(&msg).await
-                                                                {
-                                                                    tracing::warn!(
-                                                                        %node_id,
-                                                                        %daemon_id,
-                                                                        "param persisted in store; runtime forwarding is best-effort and failed: {e}"
-                                                                    );
-                                                                }
-                                                    } else {
-                                                        // Dataflow may have been removed after target resolution.
-                                                        tracing::warn!(
-                                                            %dataflow_id,
-                                                            %node_id,
-                                                            "param persisted in store; running dataflow disappeared before runtime forwarding"
-                                                        );
-                                                    }
-                                                }
-                                                Ok(ControlRequestReply::ParamSet)
-                                            }
-                                            Err(e) => Err(e),
-                                        }
-                                    }
-                                    Err(e) => Err(eyre!("failed to serialize param value: {e}")),
-                                },
-                                Err(e) => Err(e),
-                            };
+                                if let ParamTarget::Running { daemon_id } = target {
+                                    let df = running_dataflows.get_mut(&dataflow_id).ok_or_else(
+                                        || {
+                                            eyre!(
+                                                "param persisted in store but running dataflow `{dataflow_id}` disappeared before runtime forwarding for node `{node_id}`"
+                                            )
+                                        },
+                                    )?;
+                                    df.append_state_log(StateCatchUpOperation::SetParam {
+                                        node_id: node_id.clone(),
+                                        key: key.clone(),
+                                        value: value.clone(),
+                                    });
+
+                                    let msg = serde_json::to_vec(&Timestamped {
+                                        inner: DaemonCoordinatorEvent::SetParam {
+                                            dataflow_id,
+                                            node_id: node_id.clone(),
+                                            key: key.clone(),
+                                            value: value.clone(),
+                                        },
+                                        timestamp: clock.new_timestamp(),
+                                    })
+                                    .map_err(|e| {
+                                        eyre!("failed to serialize SetParam event for node `{node_id}`: {e}")
+                                    })?;
+
+                                    let conn =
+                                        daemon_connections.get_mut(&daemon_id).ok_or_else(|| {
+                                            eyre!(
+                                                "param persisted in store but daemon `{daemon_id}` is not connected"
+                                            )
+                                        })?;
+                                    let reply_raw = conn.send_and_receive(&msg).await.map_err(|e| {
+                                        eyre!(
+                                            "failed to forward SetParam to daemon `{daemon_id}` for node `{node_id}`: {e}"
+                                        )
+                                    })?;
+                                    ensure_set_param_forward_applied(&reply_raw, &node_id)?;
+                                }
+                                Ok(ControlRequestReply::ParamSet)
+                            }
+                            .await;
                             let _ = reply_sender.send(reply);
                         }
                         ControlRequest::DeleteParam {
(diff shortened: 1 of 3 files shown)


Bhanudahiyaa:
@heyong4725 maintainer context:

  This PR tightens correctness for running-target param mutations by aligning control-plane success with runtime-delivery success.

  ### Key behavior change
  For running targets:

  - daemon returns error if param event cannot be delivered (`not connected`, `channel full`, `channel closed`)
  - coordinator requires daemon `SetParamResult/DeleteParamResult(Ok(()))` before returning `ParamSet/ParamDeleted`
  - daemon rejection or unexpected reply is surfaced as `ControlRequestReply::Error(...)`

  ### Why this scope
  This is intentionally a narrow, independent fix at the CLI/coordinator/daemon boundary:
  - no protocol schema changes
  - no store backend changes
  - no unrelated refactors

  ### E2E note
  Strict semantics surfaced transient startup timing in param E2E (node channel readiness). I added bounded retry in E2E helper to keep tests deterministic while preserving strict runtime semantics.

  ### Validation rerun
  - `cargo fmt --all -- --check`
  - `cargo clippy -p dora-daemon -p dora-coordinator -- -D warnings`
  - `cargo test -p dora-daemon -- --nocapture`
  - `cargo test -p dora-coordinator -- --nocapture`
  - `cargo test --test ws-cli-e2e e2e_param_ -- --nocapture --test-threads=1`

  All pass on latest upstream-based branch.

heyong4725:
Thanks for this PR @Bhanudahiyaa. Reviewed the diff against main.

## Verdict: merge-worthy

The contract improvement is real and the code is clean. Silently succeeding when delivery didn't happen was a genuine correctness gap.

## What's good

- **Clean extraction** of `deliver_param_update_strict` / `deliver_param_delete_strict` helpers (`binaries/daemon/src/lib.rs:200-237`)
- **Coordinator refactor** from deeply nested match/match/match pyramids into `async { ... }.await` blocks with `?` propagation is a nice readability win
- **Reply validators** (`ensure_set_param_forward_applied` / `ensure_delete_param_forward_applied`) properly handle the error + unexpected-variant cases
- **Targeted tests** for the new failure modes (not connected, channel full, rejected daemon reply)
- **E2E retry helper** (`set_param_with_retry`) demonstrates the correct client-side pattern for the new strict semantics

## Two small asks before merge

1. **Asymmetric strictness in StateCatchUp.** The `StateCatchUpOperation::SetParam`/`DeleteParam` path in `binaries/daemon/src/lib.rs:1816-1859` still silently drops on channel full/closed (`Ok(false) => {}`). Either apply the same strict semantics there too, **or** add a one-line comment explaining why catch-up is intentionally lenient (catch-up is recovery, not a user-facing API, so lenience is defensible).

2. **"Eventually consistent" property isn't documented.** In the coordinator, `store.put_node_param` runs *before* runtime forwarding. When delivery fails:
   - Store has the value ✓
   - State log has the entry ✓
   - Runtime hasn't received it yet
   - User sees 'error'
   - Next catch-up delivers it

   Functionally fine (retries are idempotent, the new retry helper handles it), but the PR description says 'non-delivery paths now return errors' which implies the value wasn't persisted. Worth adding a sentence: 'Note: the value is still persisted in the store before runtime delivery is attempted, so a failed SetParam will be delivered on the next catch-up. The error reflects synchronous delivery status only.'

## CI failures are pre-existing

The 3 failing jobs (Examples ubuntu, CLI Tests ubuntu/macos) are unrelated — smoke test infrastructure issues (fixed in #212) and PyPI version mismatch from dora-hub packages. Not caused by this PR.

Happy to merge once either #1 is addressed or you confirm the catch-up asymmetry is intentional.

heyong4725:
## Re-review of PR #196

Focused pass on the strict/lenient split against the current branch state.

## Verdict: REQUEST CHANGES

The architectural direction is sound. Strict delivery for running nodes + lenient catch-up for other states is the right shape. Two issues undermine the core invariant and need addressing before merge.

## HIGH

### 1. `resolve_param_target` fires strict too early — startup race with node connect

`binaries/coordinator/src/state.rs:234-240` returns `ParamTarget::Running` the moment the dataflow exists in `running_dataflows` AND the node is in `node_to_daemon`. But `node_to_daemon` is populated at spawn time (`run/mod.rs:122`), not when the node actually connects. On the daemon side, `subscribe_channels` is only populated when the node calls subscribe (`lib.rs:3077`).

Between dataflow spawn and first node connection, there's a window where:
- Coordinator sees `ParamTarget::Running` → sends strict `SetParam`
- Daemon replies `"node {id} not connected"` → caller gets `Error`
- Param is still persisted via `append_state_log` → catches up eventually on next connect

Evidence this race is real: `set_param_with_retry` helper at `tests/ws-cli-e2e.rs:558-593` silently retries on "not connected", "channel full", and "failed to apply SetParam". The helper papers over the exact race the strict invariant is supposed to expose. Test passes, behavior ships, but a user hitting the same window sees a misleading error.

**Fix:** add a `ParamTarget::SpawningNotYetConnected { daemon_id }` variant and classify nodes as strict-eligible only after the daemon has reported `NodeStatus::Running` for them. Or expose a retryable-vs-terminal distinction in the strict error type so the CLI can surface "not yet connected, try again" differently from "permanent failure."

### 2. `set_param_with_retry` masks the signal instead of testing it

`tests/ws-cli-e2e.rs:558-593` — the retry loop silently absorbs three error classes including `"failed to apply SetParam"`. No E2E test asserts that a *non-transient* strict failure (e.g., node exists in `node_to_daemon` but its channel was permanently closed, or wrong node ID) propagates all the way through as a non-retryable error.

The core PR claim — "strict errors propagate to callers" — is verified only in unit tests (`daemon/src/lib.rs:4693, 4745`, `coordinator/src/lib.rs:2984, 2997`), not in any integration scenario. That leaves the caller-facing contract untested.

**Fix:** add one E2E case that provokes a permanent strict failure (e.g. a node ID that exists in `node_to_daemon` but is known disconnected) and asserts the error response is returned immediately — no retry loop — with a clear message.

## MEDIUM

### 3. State log appended before strict-delivery outcome is known

`binaries/coordinator/src/lib.rs:1192-1222` — `df.append_state_log(...)` runs before `conn.send_and_receive(&msg)`. Intended behavior: param is durable even if send fails, catches up on reconnect. But the user-visible result is "error returned to caller + silent eventual delivery on next catch-up." A strict API that returns an error while the value eventually applies is surprising.

Either document this explicitly as the intended behavior (the comment at 1179-1181 gestures at it but doesn't state the user-visible consequence), or move `append_state_log` to after successful strict delivery so error returns and state are consistent.

### 4. Stopping nodes still classified as Running

`state.rs:234-240` doesn't check node-level state. A node that's stopping (received `NodeEvent::Stop` but not yet removed from `subscribe_channels`) races with param send. Same misleading "channel closed" error as #1 above.

## LOW

### 5. Strict-delivery error messages omit the param key

`daemon/src/lib.rs:210, 216, 217` — errors say "node {id} not connected" / "channel full" / "channel closed" without the failing key name. Coord wrap at `lib.rs:2219` adds node ID but still not the key. Low-effort fix, improves debuggability significantly.

### 6. No explicit test for state-log-before-send partial-failure case

No unit test asserts that if `send_and_receive` returns an error after `append_state_log`, the log entry is correctly present and replays on next catch-up. The happy path is implicit in integration; the partial-failure healing path isn't directly asserted.

## What's good

- Unit test coverage for new code paths is solid on both sides.
- The strict/lenient split is the right shape architecturally.
- Catch-up leniency documentation (e833cc17) clarifies semantics that were ambiguous.

## Before merge

Items 1 and 2 are the actual blockers — they mean the strict claim is effectively untested and frequently fails for benign reasons. Items 3-6 are follow-ups.

Correction from earlier review: I initially flagged `lib.rs:2097` as a malformed comment. False alarm — that line is a valid `//` comment. Disregard.

heyong4725 (commented):
I think there is still a correctness hole in the new strict-delivery flow.

The coordinator now persists first and returns an error when the live forward fails, with the expectation that the mutation will still be replayed later from the state log (`binaries/coordinator/src/lib.rs`, around the new `SetParam` / `DeleteParam` paths).

But the daemon `StateCatchUp` path is still best-effort: when replay hits a missing node channel or `Ok(false)` from `send_with_timestamp`, it keeps going and then unconditionally sends `StateCatchUpAck { ack_sequence: max_seq }` (`binaries/daemon/src/lib.rs`, around the `StateCatchUp` handler). The coordinator then accepts that ack and prunes the log (`binaries/coordinator/src/lib.rs`, `Event::DaemonStateCatchUpAck`).

That means a mutation that initially failed with `not connected` / `channel full` can be replayed later, dropped again during catch-up, acked anyway, and then removed from the replay log permanently even though the runtime never received it. In other words, the new strict direct-forwarding path still depends on a replay path that can silently lose the update.

I think the fix needs one of these:
1. make catch-up ack only what was actually deliverable/applied, or
2. make catch-up strict for param mutations and avoid advancing/pruning ack on partial failure.

I also don't see coverage for this path in the added tests. The new tests validate the direct forward validators/helpers, but not the replay + ack interaction that now carries the recovery guarantee.

heyong4725:
## /review on PR #196

**STATUS: DONE_WITH_CONCERNS — partially superseded; recommend close-and-reopen with a rebased, scoped submission.**

### Where things stand

- Prior review [on 2026-04-16](https://github.com/dora-rs/adora/pull/196#pullrequestreview-4123935620) flagged a correctness hole: strict direct-forward semantics rely on state-log replay, but daemon `StateCatchUp` was best-effort — it would silently drop failures during replay and then unconditionally ack `max_seq`, pruning the log.
- Two options were suggested: (1) ack catch-up only through what was actually applied, or (2) strict catch-up for param mutations.
- **Option 1 landed on main in [`1fbe91d7d`](https://github.com/dora-rs/adora/commit/1fbe91d7d) at 09:41 today** (`fix(state): ack catch-up only through delivered entries`). Main now tracks `applied_through` and only acks up to that point; the state log retains entries that weren't deliverable, so they get re-tried on the next catch-up instead of being silently pruned.
- PR #196 hasn't been updated since 2026-04-14. It's **38 commits behind main**.
- `git merge --no-ff pr-196` onto main **conflicts** in `binaries/daemon/src/lib.rs` (both changes touch the same param-delivery and catch-up regions).
- CI is red on 7 jobs (Cross-check aarch64 + i686, Examples on all 3 platforms, CLI Tests Windows, ROS2 Bridge Examples) — partially due to staleness.

### What PR #196 still uniquely contributes

With the catch-up fix on main, the system is now self-healing — failed direct forwards eventually get replayed correctly and never get acked away. But **user-visible semantics** are still wrong without this PR's other half:

- A user running `dora set-param ...` gets `Ok` even when the runtime node wasn't connected or the channel was full.
- The inconsistency heals on the next reconnect-triggered catch-up, which may be arbitrarily far in the future.
- Without this PR's strict direct-forward semantics, the user has no signal to retry or investigate.

So the *conceptual* work in this PR (strict direct forwarding at daemon + coordinator layers) is still valuable **on top of** `1fbe91d7d`. Both layers together close the loop.

### Why I'd close rather than rebase-in-place

1. **Author hasn't responded to the prior review in 3 days** — PR is stale by project convention.
2. **Main's fix reshaped `StateCatchUp` handling**, which is exactly the region the PR's conflict lives in. A good rebase isn't just "fix the merge" — it's a re-audit of whether the two fixes compose correctly, which will be easier to review as a fresh PR.
3. **The PR's scope shrank** after `1fbe91d7d` landed. Some of the original motivation ("replay path that can silently lose updates") is now moot — that part was the catch-up fix, which is already in. A rebased submission should focus purely on the strict direct-forward layer.
4. **CI is red across 7 jobs** and the author hasn't investigated. A fresh rebased submission would test cleanly.

### Recommendation

**Close with a gracious note crediting the contribution.** Body should:

- Thank Bhanu for surfacing the correctness gap — the review thread directly motivated the catch-up fix that landed in `1fbe91d7d`.
- Explain that one half of the fix is already in via `1fbe91d7d`.
- Invite a rebased, smaller PR that focuses just on the strict direct-forward layer (daemon `deliver_param_update_strict` / `deliver_param_delete_strict` + coordinator reply validators), now composing with the catch-up fix on main.
- Note that the test coverage gap flagged in the earlier review (no test for replay + ack interaction) is now partially covered by the tests that landed with `1fbe91d7d`, but a rebased submission should still test the composition end-to-end.

### Alternative: request rebase in place

If you'd rather keep the PR thread open and let Bhanu push a rebase, that works too. The rebase is non-trivial but doable. Just note that the conflict region overlaps with the `StateCatchUp` refactor in `1fbe91d7d`, so "resolve the merge" here is closer to "re-design the direct-forward layer to compose with the new catch-up semantics". Worth the contributor-engagement if they're around.

### Unrelated

Same Typos blocker on main (`mis-read` in `docs/plan-consistency-sweep-2026-04-17.md:221`) is probably contributing to some of this PR's CI red. I'll open that 1-line typo fix separately today so it stops bleeding into contributor-PR reviews.

### Summary

| Item | Status |
|---|---|
| Prior review (catch-up gap) | Addressed independently in `1fbe91d7d` on main (option 1 of my two suggestions) |
| PR changes still valuable? | Yes — strict direct-forward is complementary to catch-up fix, and user-visible semantics still need it |
| Mergeable as-is? | No — conflicts + stale + 7 CI jobs failing |
| Author activity | No response in 3 days |
| Recommend | **Close with gracious note; invite rebased, scoped resubmission** |

heyong4725:
Closing per the /review above — PR is 38 commits stale, conflicts with [`1fbe91d7d`](https://github.com/dora-rs/adora/commit/1fbe91d7d) on main, and CI is red.

**Thank you @Bhanudahiyaa.** The correctness gap you surfaced ([`review comment 2026-04-16`](https://github.com/dora-rs/adora/pull/196#pullrequestreview-4123935620) — "param mutation reports success even when runtime delivery didn't happen") directly motivated the catch-up fix that landed on main today as `1fbe91d7d` (`fix(state): ack catch-up only through delivered entries`). Your PR's half of the problem (strict catch-up ack) is now resolved independently.

The other half you proposed — strict direct-forward semantics at the daemon + coordinator layers — is still valuable. Closing here because:

1. Rebasing onto main isn't mechanical: the conflict overlaps with the `StateCatchUp` refactor in `1fbe91d7d`. A rebased version needs to re-audit how strict direct-forward composes with the new catch-up semantics.
2. No updates to the PR in 3 days.
3. The scope shrank after `1fbe91d7d` landed — a fresh submission can focus purely on the direct-forward layer and compose cleanly with the catch-up fix already in main.

We're going to carry forward the strict direct-forward part as a separate, smaller follow-up commit — I'll attribute your original commit in the message so your contribution is credited in history. If you'd like to pick up other work, epic #287 has open items for dora 1.0 consolidation that we'd welcome help on.

PR #200, opened 2026-04-13
fix(coordinator): replace prometheus with OpenTelemetry metrics bridge

### LINKED ISSUE #197: Isn't using Prometheus on coordinator a regression from having the current stack in open telemetry
From the readme it says that coordinato use Prometheus creating a vendor lockin.

heyong4725:
## Summary

Closes #197.

- Replace direct `prometheus` crate usage in coordinator with OpenTelemetry `Meter` API
- Bridge to Prometheus text format via `opentelemetry-prometheus` for the `/metrics` scrape endpoint
- Metrics now flow through the standard OTel pipeline (consistent with daemon/runtime) while the scrape endpoint still works for Prometheus users

### What changed

| File | Change |
|------|--------|
| `binaries/coordinator/Cargo.toml` | Added `opentelemetry`, `opentelemetry_sdk`, `opentelemetry-prometheus` as optional deps under the `prometheus` feature |
| `binaries/coordinator/src/prometheus_metrics.rs` | Rewrote to use OTel `Meter` + `Gauge` instruments instead of raw `prometheus` `GaugeVec`. Registry created via `opentelemetry-prometheus` bridge. `SharedMetrics` is now `Arc<Metrics>` (no `Mutex` needed) |
| `binaries/coordinator/src/lib.rs` | Updated `NodeMetrics` event handler to use OTel `gauge.record()` with `KeyValue` attributes instead of `with_label_values().set()` |

### Why this matters

The coordinator was the only component using the `prometheus` crate directly, bypassing the existing OpenTelemetry stack in `libraries/extensions/telemetry/`. This created vendor lock-in and an inconsistent observability story. Now all components use OTel-native metrics.

## Test plan

- [x] `cargo build -p dora-coordinator` (without prometheus feature)
- [x] `cargo build -p dora-coordinator --features prometheus`
- [x] `cargo test -p dora-coordinator` — all 50 tests pass
- [x] `cargo clippy -p dora-coordinator --features prometheus -- -D warnings`


heyong4725:
Review result: no findings. I checked the coordinator metrics rewrite in `binaries/coordinator/Cargo.toml`, `binaries/coordinator/src/lib.rs`, and `binaries/coordinator/src/prometheus_metrics.rs` with focus on whether switching from `prometheus::GaugeVec` updates to OTel `Gauge::record()` changes `/metrics` behavior. The Prometheus bridge uses cumulative LastValue aggregation for gauges, so the existing latest-value-per-label-set semantics are preserved.

Residual risk: there is still no coordinator regression test that asserts `/metrics` output with the `prometheus` feature enabled, so this is validated by build/test coverage rather than a direct scrape-format test.

Validation run: `cargo test -p dora-coordinator --features prometheus`

haixuanTao:
I think it does not adress the core issue as it still binds to prometheus, with opentelemetry as support but we want to export to opentelemetry. It would also be nice to not have yet another abstraction to worry about. Can coordinator just use the default dora-metrics?


Is there any way we can test this? You're mentioning prometheus but there is no readme, there is no getting started, there is no mention of configuration?

heyong4725:
Thanks for the feedback @haixuanTao. Reworked the implementation:

- **Removed the prometheus scrape endpoint entirely** — no more `/metrics` HTTP handler, no `prometheus` or `opentelemetry-prometheus` crate deps
- **Using the global OTel meter from `dora-metrics`** — coordinator now records gauges through the same OTLP pipeline as daemon/runtime
- **Renamed feature from `prometheus` to `metrics`** — consistent naming with daemon
- **Dropped the `Mutex`** — OTel gauges are thread-safe

Users who want Prometheus can point an OTel Collector at the OTLP endpoint with a prometheus exporter, which is the standard approach.

The coordinator no longer maintains its own metrics abstraction — it just creates gauges via `opentelemetry::global::meter("dora-coordinator")` and records values.

heyong4725:
Review findings:

1. `binaries/coordinator/src/lib.rs:196` initializes a fresh `SdkMeterProvider`, but `binaries/coordinator/src/otel_metrics.rs:23` creates instruments from `opentelemetry::global::meter("dora-coordinator")`. `dora_metrics::init_metrics()` only returns a provider; it does not install it globally. Without `opentelemetry::global::set_meter_provider(...)`, these new coordinator gauges are recorded against the default no-op global provider, so the PR drops coordinator metrics entirely even when the `metrics` feature is enabled.

2. This branch removes the coordinator `prometheus` feature and `/metrics` endpoint (`binaries/coordinator/src/ws_server.rs`, `binaries/coordinator/Cargo.toml`), but the user-facing contract still advertises that feature and endpoint in `README.md:114` and `guide/src/getting-started/installation.md:46`. That is a breaking behavior change, not just an internal refactor, and it is undocumented in the PR. Users enabling `prometheus` or scraping `/metrics` from the coordinator would regress immediately.

Validation: reviewed the current PR head diff and coordinator metrics wiring; also started `cargo test -p dora-coordinator --features metrics` on the branch while checking the source paths above.

PR #204, opened 2026-04-13
feat(types): complete type annotations with nested refs and List<T>

### LINKED ISSUE #199: Type annotation documentation miss a lot of key details
https://github.com/dora-rs/adora/blob/main/docs/types.md 

The logic is very loose as struct can be composed of many different types as well as nan. None of it is documented rigorously

heyong4725:
## Summary

Closes #199.

The type annotation system had two gaps: 7 struct types lacked field definitions (silently skipping schema validation), and the YAML format only supported primitive Arrow types as field types — no way to express nested structs or lists.

This PR addresses both:

### 1. Extend type system for nested refs + List types

- **`resolve_field_type()`** — new function in `types.rs` that resolves field type strings supporting:
  - Primitive Arrow types (`Float64`, `Utf8`, etc.)
  - `List<T>` syntax (mapped to `LargeList`) — works with any inner type
  - Struct type references by short name (`Vector3`) or full URN (`std/math/v1/Vector3`)
  - Depth-limited to 8 to prevent circular reference panics

- **`resolve_short_name()`** on `TypeRegistry` — finds types by unique URN suffix match (e.g., `Vector3` → `std/math/v1/Vector3`). Returns `None` if ambiguous.

- **`to_arrow_schema_with_registry()`** on `TypeDef` — registry-aware schema builder that uses `resolve_field_type`. The existing `to_arrow_schema()` is unchanged.

- **`check_schema_compat()`** in `validate.rs` — updated to use registry-aware method, so nested struct types are now validated at the field level.

### 2. Complete all 7 incomplete struct type definitions

| Type | Fields added |
|------|-------------|
| **Pose** | position (Vector3), orientation (Quaternion) |
| **Transform** | translation (Vector3), rotation (Quaternion) |
| **Twist** | linear (Vector3), angular (Vector3) |
| **JointState** | names (List\<Utf8\>), positions/velocities/efforts (List\<Float64\>) |
| **Odometry** | pose (Pose), twist (Twist), frame_id (Utf8) |
| **Detection** | boxes (List\<BoundingBox\>), count (UInt32) |
| **Segmentation** | width, height, num_classes (UInt32), mask (LargeBinary), class_names (List\<Utf8\>) |

### 3. Documentation updates

- Added "Nested Type References" and "List Types" sections to `docs/types.md`
- Updated all Standard Type Library tables with complete field definitions

### Files changed

| File | Change |
|------|--------|
| `libraries/core/src/types.rs` | `resolve_field_type()`, `resolve_short_name()`, `to_arrow_schema_with_registry()` + 16 new tests |
| `libraries/core/src/descriptor/validate.rs` | Use registry-aware schema method in `check_schema_compat()` |
| `types/std/math/v1.yml` | Pose, Transform field definitions |
| `types/std/control/v1.yml` | Twist, JointState, Odometry field definitions |
| `types/std/vision/v1.yml` | Detection, Segmentation field definitions |
| `docs/types.md` | New sections + updated type tables |

## Test plan

- [x] `cargo test -p dora-core` — 140 tests pass (16 new)
- [x] `cargo clippy -p dora-core -- -D warnings` — clean
- [x] New test `all_std_struct_types_have_schemas` verifies all 13 struct types resolve schemas

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -83,6 +83,42 @@ input_types:
   audio: std/media/v1/AudioFrame[sample_type=i16]
 ```
 
+### Nested Type References
+
+Struct field types can reference other defined types by short name or full URN:
+
+```yaml
+types:
+  MyPose:
+    arrow: Struct
+    fields:
+      - name: position
+        type: Vector3          # resolved to std/math/v1/Vector3
+      - name: orientation
+        type: Quaternion       # resolved to std/math/v1/Quaternion
+```
+
+Short names (e.g., `Vector3`) are resolved by unique suffix match across all registered types. If a short name is ambiguous (multiple types share it), use the full URN (e.g., `std/math/v1/Vector3`).
+
+### List Types
+
+Fields can use `List<T>` syntax to express typed lists. These are mapped to Arrow `LargeList`:
+
+```yaml
+types:
+  MyDetection:
+    arrow: Struct
+    fields:
+      - name: boxes
+        type: List<BoundingBox>   # LargeList of BoundingBox structs
+      - name: scores
+        type: List<Float32>       # LargeList of Float32
+      - name: labels
+        type: List<Utf8>          # LargeList of strings
+```
+
+`List<T>` supports any inner type: primitives, struct references, or nested lists.
+
 ## Standard Type Library
 
 ### `std/core/v1`
@@ -106,16 +142,16 @@ input_types:
 |------|-----------|--------|-------------|
 | `Vector3` | Struct | x, y, z (Float64) | 3D vector |
 | `Quaternion` | Struct | x, y, z, w (Float64) | Quaternion |
-| `Pose` | Struct | position, orientation | 6-DOF pose |
-| `Transform` | Struct | translation, rotation | Coordinate transform |
+| `Pose` | Struct | position (Vector3), orientation (Quaternion) | 6-DOF pose |
+| `Transform` | Struct | translation (Vector3), rotation (Quaternion) | Coordinate transform |
 
 ### `std/control/v1`
 
-| Type | Arrow Type | Description |
-|------|-----------|-------------|
-| `Twist` | Struct | Linear and angular velocity |
-| `JointState` | Struct | Joint positions, velocities, efforts |
-| `Odometry` | Struct | Pose + Twist in a reference frame |
+| Type | Arrow Type | Fields | Description |
+|------|-----------|--------|-------------|
+| `Twist` | Struct | linear (Vector3), angular (Vector3) | Linear and angular velocity |
+| `JointState` | Struct | names (List\<Utf8\>), positions/velocities/efforts (List\<Float64\>) | Joint state |
+| `Odometry` | Struct | pose (Pose), twist (Twist), frame_id (Utf8) | Pose + Twist in a reference frame |
 
 ### `std/media/v1`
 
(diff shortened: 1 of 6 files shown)


heyong4725:
Review finding:

`libraries/core/src/types.rs:86` and `libraries/core/src/types.rs:160` still build schemas with `filter_map`, so unresolved nested field types are silently dropped instead of failing schema construction. That was already lenient for primitive-only schemas, but this PR makes it more dangerous because `check_schema_compat()` now relies on these registry-aware schemas for nested validation (`libraries/core/src/descriptor/validate.rs:1105`).

If a field uses an ambiguous short name or a misspelled nested type, the consumer schema can collapse to a partial struct and `schema_compatible()` may incorrectly accept an upstream type that is actually missing required nested fields. This should fail closed: if any declared field cannot be resolved, schema building should return `None` or an error, not a truncated schema.

Validation run: `cargo test -p dora-core`

heyong4725:
Non-blocking follow-up improvements:

- Make schema construction fail closed by replacing the `filter_map` paths with an all-or-nothing collect, ideally returning a structured error that includes the unresolved field name and type string.
- Add a regression test where a nested field references an unknown or ambiguous short name and assert that validation reports an incompatibility instead of accepting a partial schema.
- Add one targeted test for parameterized struct types flowing through `to_arrow_schema_with_registry`, even if current stdlib examples do not use nested parameterized fields yet.
- Consider preferring full URNs in the built-in YAML files and docs, while keeping short-name support for users. That reduces future ambiguity as the type library grows.

heyong4725:
Follow-up: I prototyped the fail-closed fix in an isolated PR 204 worktree and validated it.

What changed in the prototype:
- Replaced the `filter_map` schema-building paths in `libraries/core/src/types.rs` with all-or-nothing collection, so unresolved nested field types now cause schema construction to return `None` instead of silently dropping fields.
- Added regression tests for:
  - unknown nested type references
  - ambiguous nested short-name references

Validation run:
- `cargo fmt --all -- --check`
- `cargo test -p dora-core`

Result: both passed, and the new tests catch the partial-schema acceptance case discussed in review.

heyong4725:
Addressed both review comments in 4f5ce1c7:

### Fail-closed schema construction

`build_schema` now uses all-or-nothing `collect::<Option<Vec<_>>>()` instead of `filter_map`. If any declared field can't resolve (misspelled type, ambiguous short name, depth exceeded), the entire schema returns `None`. `check_schema_compat` then skips rather than validating against a truncated schema.

### Regression tests added

- `unresolvable_nested_field_returns_none` — struct with one good + one unknown field → `None`
- `ambiguous_short_name_in_nested_field_returns_none` — duplicate Vector3 → short name ambiguous → `None`
- `parameterized_struct_schema_resolves` — `AudioFrame[sample_type=f32]` resolves schema through registry

### On full URNs in built-in YAML

Keeping short names in the YAML for now — they're unambiguous within the std library and more readable. If the type library grows and collisions appear, we can switch to full URNs then.

All 145 tests pass.

haixuanTao:
So, one of the design of dora was to be able to accept generics as inputs, because by design it's very hard to tell if data is going to be u8 or u16 or f32 ... , same thing for representation, you could have quaternion you could have angle, you could have bounding box in xywh or xyxy, you could have joint space or cartesian space for pose. You can clearly see within ros documentation that there is no clear unique definition for a lot of things. Which make dora using arrow kind of powerful but so having a typing system need to also be able to accept generics imo. You could check dora-rerun for more.

also, structure in arrow is actually very not easy to use the api is a bit frustrating especially in python hence me avoiding it.

haixuanTao:
Typically it's hard to mix float with string in a struct in python

heyong4725:
Good points @haixuanTao. The type annotation system is fully opt-in and designed with exactly this flexibility in mind:

1. **No annotations = no enforcement.** If a node declares no `output_types`/`input_types`, any Arrow data passes through freely. This is the default.

2. **Warnings only by default.** Even when types are declared, mismatches produce warnings, not errors. Users opt into strict checking with `strict_types: true` or `--strict-types`.

3. **Generics are supported.** Nodes can accept any data shape without declaring types. The annotation system is for users who \*want\* to catch mismatches early (e.g., plugging a BoundingBox output into an Image input).

So the design accommodates both worlds:
- Robotics research where data representations vary (quaternion vs angle, xywh vs xyxy) — just don't annotate, or use a common base type
- Production pipelines where catching type mismatches at build time saves debugging time

The Arrow struct API friction in Python is a real concern but orthogonal to this PR — the type annotations live in the YAML layer and don't affect how nodes construct or consume Arrow data at runtime. A node sending raw `pa.array([1.0, 2.0, 3.0])` works the same whether or not a type annotation exists.

Issue #215, opened 2026-04-14
CI coverage gaps: ~80% of CLI commands and ~60% of advertised features are not exercised


haixuanTao:
## Summary

Audit of [README.md](README.md) advertised functionality vs. `.github/workflows/ci.yml` coverage. CI covers the core end-to-end path (run/up/start/stop, Rust/Python/C/C++ templates, basic examples, ROS2 bridge, fault-tolerance e2e, benchmark, audit, unwrap-budget, cross-check, MSRV), but a large fraction of the surface area the README promises is never exercised in CI.

Rough untested percentages (see breakdown below):

| Surface | Total items in README | Exercised in CI | Untested |
|---|---|---|---|
| CLI commands | 47 | ~9 | **~81%** |
| README example dataflows | ~33 | ~12 | **~64%** |
| Feature bullets (Performance / DX / Prod / Observability / Ecosystem) | ~23 | ~9 | **~61%** |
| Platform matrix (C/C++ templates) | 3 platforms | 1 (Linux only) | **~67%** |

### Lifecycle (2/7 untested)
- `dora restart <ID>`
- `dora down` is invoked as `dora destroy` in one place; `dora down` itself is not tested

### Monitoring (21/23 untested)
- `dora logs <ID>`
- `dora top` / `dora inspect top`
- `dora topic list`, `dora topic hz`, `dora topic echo`, `dora topic info`, `dora topic pub`
- `dora node list`, `dora node info`, `dora node add`, `dora node remove`, `dora node connect`, `dora node disconnect`, `dora node restart`, `dora node stop`
- `dora param list`, `dora param get`, `dora param set`, `dora param delete`
- `dora trace list`, `dora trace view`
- `dora record <PATH>`, `dora replay <FILE>`

### Cluster management (7/7 untested — 0% coverage)
- `dora cluster up`, `dora cluster status`, `dora cluster down`, `dora cluster install`, `dora cluster uninstall`, `dora cluster upgrade`, `dora cluster restart`

### Setup and utilities (9/10 untested)
- `dora doctor`
- `dora status` / `dora check`
- `dora graph` (Mermaid/HTML visualization)
- `dora expand` (module flattening)
- `dora validate` and `dora validate --strict-types`
- `dora system`
- `dora completion`
- `dora self update`

Only `dora new` (Rust/Python/C/C++ templates) is exercised.

## README features advertised but not verified by CI

[...]

PR #220, opened 2026-04-14
ci(examples): remove unused CLI install (-45 runner-minutes/PR)

heyong4725:
## Summary

Refs #215. Removes an unused `cargo install --path binaries/cli --locked` step from the Examples CI job. Reclaims ~45 runner-minutes per PR run.

## What I found

The Examples job had a "Build CLI" step that installed `dora` to `~/.cargo/bin`:
```yaml
- name: Build CLI
  timeout-minutes: 45
  run: cargo install --path binaries/cli --locked
```

But every example in this job calls into `dora_cli` **as a library** via `examples/*/run.rs`:
```rust
use dora_cli::{build, run};
...
fn main() {
    build(...)?;
    run(...)?;
}
```

Verified across all 9 `run.rs` files in `examples/`: **zero** shell out to the `dora` binary. Grep `Command::new("dora")` returns no matches.

So the CLI install was pure waste — compiled once, installed to PATH, never invoked.

## Impact

| | Before | After |
|---|---|---|
| Examples job — ubuntu | ~60 min | ~45 min |
| Examples job — macos | ~60 min | ~45 min |
| Examples job — windows | ~60 min | ~45 min |
| **Runner-minutes per PR** | ~180 | ~135 |

Saves ~45 min of runner time per PR, ~15 min of wall-clock on the slowest platform.

## Risk

Low. This is a pure deletion of an unused step. The example nodes are still built via `cargo build --examples`, which links `dora_cli` into each example binary.

If this is wrong and some example does need the binary, the job will fail loudly on that step — no silent regression.

## Not in this PR

The `cli` job (separate) **does** use the `dora` binary via shell and keeps its `cargo install` step. Pre-building CLI artifacts for the `cli` job and for the nightly jobs is a separate optimization (separate PR to follow).

## Test plan

- [x] Grepped all `examples/*/run.rs` for `Command::new("dora")` or `"dora"` shell-out — none found
- [x] YAML validates: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"`
- [x] Examples still link `dora_cli` via `Cargo.toml` dependencies (not via PATH)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,9 +129,9 @@ jobs:
           Remove-Item "C:\Android" -Force -Recurse
           Remove-Item "C:\hostedtoolcache\windows\go" -Force -Recurse
           Remove-Item "C:\hostedtoolcache\windows\CodeQL" -Force -Recurse
-      - name: Build CLI
-        timeout-minutes: 45
-        run: cargo install --path binaries/cli --locked
+      # No `cargo install --path binaries/cli` here: every example in this
+      # job calls into `dora_cli` as a library (via `examples/*/run.rs`).
+      # The `dora` CLI binary is not on PATH and not required.
       - name: Build examples
         timeout-minutes: 30
         run: cargo build --examples


heyong4725:
No findings.

I re-checked the Examples job change against the current repo state. Removing the CLI install from `.github/workflows/ci.yml` looks safe because the Examples job runs example binaries built by `cargo build --examples`, and those examples call into `dora_cli` as a library rather than requiring a `dora` binary on PATH. I did not find any example `run.rs` that shells out to `dora`.

Validation:
- Parsed `.github/workflows/ci.yml` successfully
- Reviewed the PR diff in the Examples job
- Grepped example `run.rs` files for `dora` shell-outs / PATH lookups and only found normal tool invocations like `cargo`, `clang`, `cmake`, and `pkg-config`

The PR 219 nightly workflow issues are separate and not addressed by this PR.

PR #221, opened 2026-04-14
ci(nightly): fix Python setup + dedup issue filing (follow-up to #219)

heyong4725:
## Summary

Follow-up to #219 addressing three review findings. All valid.

### 1. `log-sinks` job missing Python setup (would fail false-positive)

All three log-sink dataflows have Python source nodes:
- `examples/log-sink-file/dataflow.yml` → `sensor.py`, `processor.py`
- `examples/log-sink-alert/dataflow.yml` → `source.py`
- `examples/log-sink-tcp/dataflow.yml` → `source.py`

The nightly job ran `dora run <yaml> --stop-after 15s` with no `uv venv`, no `apis/python/node` install, and no `--uv` flag. As the reviewer noted, this would fail on environment setup, not real regressions.

**Fix:** added `setup-python` + `setup-uv` + venv bootstrap (mirrors `ci.yml`). Passes `--uv` to each `dora run`.

### 2. Split `service-action-streaming` into two jobs

Same issue as #1 but only for the `streaming-example` step. `service-example` and `action-example` are Rust-only; `streaming-example` has three Python nodes (`prompt_source.py`, `generator.py`, `sink.py`).

**Fix:** split into two jobs. `service-action` stays lean (no Python). New `streaming` job has the Python setup.

### 3. `file-issue-on-failure` would spam issues

Original: `gh issue create ...` runs on every scheduled failure. N-day outage → N new issues.

**Fix:** check for an existing open issue with the `nightly-regression` label. If one exists, comment on it (dated). If not, create a new one. PR description noted "no PR-level noise" — this now matches that intent.

```bash
EXISTING=$(gh issue list --state open --label nightly-regression --limit 1 --json number --jq '.[0].number // empty')
if [ -n "$EXISTING" ]; then
  gh issue comment "$EXISTING" --body "$COMMENT_BODY"
else
  gh issue create --title "..." --body "$ISSUE_BODY" --label nightly-regression
fi
```

### 4. `docs/testing-matrix.md` reproducer mismatched requirements

Same issue as #1: the local-reproducer block included `dora run examples/log-sink-file/dataflow.yml --stop-after 15s` with no setup note and no `--uv`.

**Fix:** split Rust-only vs Python-backed commands, added the venv setup incantation as a comment block above the Python commands.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on the workflow
- [x] Verified each `examples/log-sink-*` and `streaming-example` has `.py` files
- [x] Verified `service-example` and `action-example` are Rust-only (no `.py`)
- [x] `gh issue list --state open --label nightly-regression` works on this repo (label exists, created in PR #219 prep)

Will trigger via `workflow_dispatch` after merge to confirm the Python-setup jobs actually pass.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -70,7 +70,7 @@ jobs:
         run: exit 1
 
   log-sinks:
-    name: Log-sink examples
+    name: Log-sink examples (Python sources)
     runs-on: ubuntu-latest
     timeout-minutes: 20
     steps:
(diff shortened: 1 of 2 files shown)


heyong4725:
Re-review focused on the PR 219 findings: no findings.

This follow-up addresses the issues I called out on #219:
- `log-sinks` now has Python/uv setup and uses `--uv`
- `streaming-example` is split into its own Python-configured job, while service/action stay Rust-only
- nightly issue filing now deduplicates by commenting on an existing open `nightly-regression` issue instead of always creating a new one
- `docs/testing-matrix.md` now distinguishes Rust-only vs Python-backed repro steps and includes the missing Python setup plus `--uv`

I also re-validated that `.github/workflows/nightly.yml` parses cleanly.

PR #222, opened 2026-04-14
ci(nightly): pre-build CLI once, share via artifact (-43 runner-min/night)

heyong4725:
## Summary

Refs #215. Tackles the nightly-CI runner-time waste: 4 downstream jobs each did `cargo install --path binaries/cli --locked` independently (~15 min each = ~60 runner-minutes burned per nightly).

## Design

Single `build-cli` job installs the CLI once to `./dist/bin/dora` and uploads as artifact (retention 1 day). Downstream jobs (`log-sinks`, `service-action`, `streaming`, `record-replay`) download the artifact and prepend its directory to `PATH`.

## Impact

| | Before | After |
|---|---|---|
| CLI builds per nightly | 4 × 15 min = **60 min** | 1 × 15 min + 4 × 30s downloads = **~17 min** |
| Total runner cost per nightly | ~90 min | ~47 min |
| Wall-clock (critical path) | ~35 min | ~30 min (build-cli gates downstream jobs) |
| Runner-minutes saved | — | **~43 min/night** |

## Also

Dropped the Rust toolchain + rust-cache from the `streaming` job. It only runs `dora run` on Python nodes — no Rust build needed. Saves another couple of minutes per run.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on the workflow
- [x] Verified `cargo install --root ./dist` produces `./dist/bin/dora`
- [x] Artifact path `dist/bin/dora` matches what `upload-artifact` points at
- [x] Downstream jobs `chmod +x` before adding to PATH (binary loses +x on download)

Will trigger via `workflow_dispatch` after merge to confirm the artifact hand-off works end-to-end.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -240,8 +240,11 @@ jobs:
   file-issue-on-failure:
     name: File issue on nightly failure
     runs-on: ubuntu-latest
-    needs: [smoke-suite, log-sinks, service-action, streaming, record-replay]
-    if: failure() && github.event_name == 'schedule'
+    needs: [build-cli, smoke-suite, log-sinks, service-action, streaming, record-replay]
+    # `always() && contains(needs.*.result, 'failure')` covers both direct
+    # failures and the case where build-cli fails and downstream jobs get
+    # skipped (plain `failure()` would miss the skipped-cascade case).
+    if: always() && contains(needs.*.result, 'failure') && github.event_name == 'schedule'
     steps:
       - uses: actions/checkout@v4
       - name: Create or update regression issue
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -4,8 +4,9 @@ name: Nightly Regression
 #
 # Tier 1 gates per docs/testing-matrix.md:
 #   - Full example-smoke suite (~11 min)
-#   - Service/action/streaming communication patterns
-#   - Log-sink example dataflows
+#   - Service/action communication patterns (Rust-only)
+#   - Streaming example (Python)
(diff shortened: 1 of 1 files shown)


heyong4725:
Review finding:

The new build-cli job is introduced in nightly.yml, but file-issue-on-failure still only depends on smoke-suite, log-sinks, service-action, streaming, and record-replay.

If build-cli fails, those downstream jobs will be skipped, and the issue-filing job will also be skipped rather than reporting the nightly failure. That breaks the workflow intent that nightly failures file a nightly-regression issue.

build-cli needs to be included in the failure-reporting dependency chain, and this job likely needs an always()-style guard so it still runs when upstream jobs are skipped due to the shared CLI build failing.

Validation: re-checked the workflow diff and re-validated that nightly.yml parses cleanly.

heyong4725:
Good catch — addressed in 6fa8fbd9.

Two fixes:
1. Added `build-cli` to the filer's `needs:` list
2. Changed `if: failure()` to `if: always() && contains(needs.*.result, 'failure')` — handles both direct-failure and skipped-cascade cases (plain `failure()` skips when cascading skips happen)

The net result: nightly failures always file/update the regression issue, whether build-cli, smoke-suite, or any downstream job is the root cause.

heyong4725:
Re-review of the current PR head: no findings.

The earlier nightly failure-reporting gap is fixed. file-issue-on-failure now includes build-cli in its needs list and uses an always() + contains(needs.*.result, failure) guard, so a shared CLI build failure will still trigger issue filing instead of silently skipping the reporter.

I also re-validated that nightly.yml parses cleanly.

Issue #228, opened 2026-04-14
qa: cluster/distributed command coverage (cluster up/status/down)


heyong4725:
**Parent:** #215

`dora cluster *` commands are 0% covered per the #215 audit. Reviewer feedback:

> Cluster/distributed operational commands still look uncovered: `dora cluster daemon`, `dora cluster coordinator`, `dora cluster destroy`, likely parts of `dora logs` / remote inspection flows.

## Scope

One nightly-only smoke test that exercises the cluster lifecycle end-to-end.

## Concrete shape

Add a new job to `.github/workflows/nightly.yml`:

```yaml
cluster-smoke:
  name: Cluster lifecycle smoke
  runs-on: ubuntu-latest
  needs: build-cli
  timeout-minutes: 15
  steps:
    - checkout + python setup + download-artifact: dora-cli + install to PATH
    - name: Cluster up
      run: dora cluster up --size 2 --wait 30s
    - name: Cluster status shows 2 daemons
      run: |
        out=$(dora cluster status --json)
        test "$(echo "$out" | jq '.daemons | length')" = "2"
    - name: Run a dataflow on the cluster
      run: dora run examples/rust-dataflow/dataflow.yml --stop-after 10s
    - name: Cluster destroy
      run: dora cluster destroy
    - name: Assert no leftover processes
      run: pgrep -f 'dora-daemon\|dora-coordinator' && exit 1 || exit 0
```

## Investigation needed first

Verify `dora cluster *` command signatures actually match this plan. From `dora cluster --help`:

```
Usage: dora cluster <COMMAND>
Commands:
  up         Start coordinator + daemons
  status     Show cluster health
  down       Tear down
  install    Install dora on remote machines
  uninstall  Remove dora from remote machines
  upgrade    Upgrade dora on remote machines
  restart    Restart the cluster
```

Check which subset makes sense for a single-host smoke test (local cluster with 2 daemon processes, no SSH/install).

## Why nightly, not PR gate

- Cluster lifecycle is heavier than the existing `multiple-daemons` example (which is in Examples job)
- Flakiness in cluster setup is more likely than in standard dataflow runs
- Nightly failure mode (issue-filing) is a good fit

## Acceptance

- `cluster-smoke` job added to `nightly.yml`
- Confirms cluster up → status → run dataflow → destroy works on a fresh runner
- Follow-up issue opened for SSH-based install/upgrade tests (requires dedicated infra, out of scope for this one)

Issue #227, opened 2026-04-14
qa: interactive/TUI CLI coverage (topic echo/hz/info/pub, top)


heyong4725:
**Parent:** #215

Interactive/TUI CLI commands are still untested per the #215 audit. Reviewer feedback:

> The biggest remaining holes are still: `dora topic echo`, `dora topic hz`, `dora topic info`, `dora topic pub`, `dora top`

## The blocker

These are TUI/interactive commands. Spawning them in CI requires either:

1. A non-interactive / machine-readable mode (e.g. `--once`, `--json`, `--stop-after N`)
2. An expect-style harness (brittle on CI runners)

Option 1 is the right answer — it's a small CLI change that makes these commands testable AND automatable by users.

## Investigation phase (this issue)

Audit which commands already have non-interactive modes, and which need new flags:

| Command | Current mode | Needed flag(s) |
|---|---|---|
| `dora top` | TUI only? Check `--once`, `--json` | TBD |
| `dora topic echo` | Streaming TUI? | `--count N`, `--timeout N` |
| `dora topic hz` | Rolling average? | `--duration N` |
| `dora topic info` | Probably already non-interactive | Maybe `--json` |
| `dora topic pub` | One-shot? | Should be straight-through |

## Implementation phase (separate PRs per command)

For each command that needs a new flag:

1. Add the non-interactive mode to the CLI (Rust change in `binaries/cli`)
2. Add Tier 0 smoke test (piggyback on `test` job)
3. Document in CLI help text

## Acceptance

- Audit output posted in this issue (which commands already work, which need flags)
- Separate PRs opened for each needed flag (grouped by command)
- Tier 0 coverage added once flags exist

## Dependencies

Must happen after sub-issue A (semantic assertions) ships — those establish the pattern we'll follow here.

heyong4725:
Update: the remaining 2/6 commands (`topic echo`, `topic hz`) are blocked by **#236** (daemon bug: `publish_all_messages_to_zenoh` doesn't actually route messages through Zenoh).

- CLI work done: #235 adds `--count`/`--duration` flags so both commands have bounded modes for scripting
- Smoke test work: can't write meaningful assertions until #236 is fixed

This issue stays open until #236 unblocks CI smoke.

PR #223, opened 2026-04-14
ci(test): add Tier 0 CLI smoke gates (closes ~20 of the 47 coverage gaps from #215)

heyong4725:
## Summary

Refs #215. Adds **Tier 0 fast CLI smoke gates** to the existing `test` job. Addresses ~20 of the 47 CLI commands flagged as untested in the audit.

## What it catches

Regressions users hit **immediately**:
- Broken argparse (commit misconfigures clap → `dora <sub> --help` crashes)
- `dora validate` silently accepting invalid YAML (schema drift)
- `dora expand` failing on module references (flattening regressions)
- `dora graph` renderer breaks (mermaid output malformed)

## Design

Piggybacks on the existing `test` job — the workspace is already compiled via `cargo build --all`, so each `cargo run --quiet -p dora-cli -- <sub>` uses the already-built binary. Each invocation is <1s.

Runs on **all 3 platforms** (ubuntu/macos/windows) — catches platform-specific argparse drift for free.

## Gates added

```yaml
- CLI smoke — argparse & help text    # 22 subcommands × --help
- CLI smoke — dora validate            # rust-dataflow + multiple-daemons
- CLI smoke — dora expand              # module-dataflow
- CLI smoke — dora graph (mermaid)     # rust-dataflow + multiple-daemons
```

The 22 subcommands: `run up down build start stop restart list logs top topic node param record replay trace doctor status new graph expand validate`. Verified locally that all accept `--help`.

## Impact

| | Before | After |
|---|---|---|
| CLI commands smoke-tested | ~9/47 (19%) | ~29/47 (62%) |
| CI time delta per platform | — | +10-20s |
| Platforms | — | 3 (ubuntu/macos/windows) |

## What's NOT in this PR

- `dora doctor` (without `--help`) — needs running coordinator. Out of scope for a test-job smoke.
- `dora top`, `dora topic echo/hz` — TUI commands, need expect-style harness.
- `dora param get/set` — covered by `ws-cli-e2e`, not duplicated here.
- `dora cluster *` — needs multi-daemon setup, belongs in Tier 2/manual.

## Test plan

- [x] Verified all 22 subcommands accept `--help` locally
- [x] Timed the smoke sequence on hot cache: 12s
- [x] YAML validates
- [x] Updated `docs/testing-matrix.md` Tier 0 table

PR #231, opened 2026-04-14
ci(test): add semantic assertions for Tier 0 CLI tests

### LINKED ISSUE #226: qa: semantic assertions for Tier 0 CLI tests (expand, graph, validate)
**Parent:** #215

Tier 0 CLI smoke tests (shipped in #223) prove commands start and parse args, but don't assert they produce correct output. Reviewer feedback on #215 called this out explicitly:

> 223 proves many commands start and parse args, but it does not prove they do the right thing. That is especially thin for: `dora graph`, `dora expand`, `dora validate --strict-types`

## Scope

Add golden-output / semantic assertions to the existing Tier 0 steps in `.github/workflows/ci.yml` → `test` job. Piggyback on the already-compiled workspace, no new CI budget.

### `dora expand` — must flatten module references

```bash
out=$(cargo run --quiet -p dora-cli -- expand examples/module-dataflow/dataflow.yml)
# Asserts inline node IDs are present
echo "$out" | grep -q 'id: sender'
echo "$out" | grep -q 'id: receiver'
# Asserts `modules:` block is gone (flattened)
echo "$out" | grep -vq '^modules:'
```

### `dora graph --mermaid` — must contain expected node labels

```bash
out=$(cargo run --quiet -p dora-cli -- graph examples/rust-dataflow/dataflow.yml --mermaid)
echo "$out" | grep -q 'rust-node'
echo "$out" | grep -q 'rust-sink'
echo "$out" | grep -q 'rust-status-node'
```

### `dora validate --strict-types` — positive + negative cases

```bash
# Positive: typed-dataflow should pass strict mode
cargo run --quiet -p dora-cli -- validate --strict-types examples/typed-dataflow/dataflow.yml

# Negative: create a fixture with an intentional type mismatch, expect non-zero exit
! cargo run --quiet -p dora-cli -- validate --strict-types tests/fixtures/type-mismatch.yml
```

Need to create `tests/fixtures/type-mismatch.yml` as part of this work.

## Why this is P1

- Directly addresses the strongest review criticism on #215
- Near-zero cost (shares existing `test` job)
- Catches the class of regressions CI currently misses: behavior drift that still produces a zero exit code

## Out of scope

- Non-Tier-0 commands (tracked elsewhere in the #215 bucket split)
- Interactive/TUI commands (separate sub-issue)

## Acceptance

- CI test job has new steps asserting semantic output for `expand`, `graph`, `validate --strict-types`
- At least one intentionally-wrong fixture for the negative `--strict-types` path
- All 3 platforms (ubuntu/macos/windows) still green

heyong4725:
## Summary

Closes #226. Addresses the strongest critique of #223 (Tier 0 CLI smoke): invocation succeeded but output content was never checked.

## What's added

Three new semantic assertion steps in the `test` job, running on all 3 platforms. Each piggybacks on the already-compiled workspace.

### `dora validate --strict-types` (positive + negative)

| Case | Fixture | Expected |
|---|---|---|
| Positive | `examples/typed-dataflow/dataflow.yml` | exit 0 |
| Negative | `tests/fixtures/type-mismatch.yml` (new) | exit != 0 |

The negative fixture declares `input_types: String` on a `Float64` output. If strict mode ever regresses to a no-op, the negative case will pass and CI will fail correctly.

### `dora expand` — must flatten modules

```bash
out=$(dora expand examples/module-dataflow/dataflow.yml)
# All 4 node IDs present
grep -q '^- id: sender' <<<"$out"
grep -q '^- id: pipeline.doubler' <<<"$out"
grep -q '^- id: pipeline.filter' <<<"$out"
grep -q '^- id: receiver' <<<"$out"
# modules: block is gone
! grep -q '^modules:' <<<"$out"
```

### `dora graph --mermaid` — contains expected nodes + flowchart header

```bash
out=$(dora graph examples/rust-dataflow/dataflow.yml --mermaid)
grep -q '^flowchart ' <<<"$out"
grep -q 'rust-node'         <<<"$out"
grep -q 'rust-sink'         <<<"$out"
grep -q 'rust-status-node'  <<<"$out"
```

## What regressions this catches

- Strict-types silently becoming a no-op (e.g. a refactor that forgets the `--strict` flag handling)
- `expand` regressing to pass-through or stopping at the module boundary
- `graph` renderer producing empty/malformed output with no error

## Test plan

- [x] All 11 assertions pass locally
- [x] Negative strict-types case exits non-zero with a clear error
- [x] YAML validates
- [x] No changes to existing assertions; only additions

## Not in this PR

- Other Tier 0 commands (`trace list/view`, etc.) — tracked in #226 follow-ups
- TUI command coverage — #227
- Cluster commands — #228

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,15 +110,60 @@ jobs:
         run: |
           cargo run --quiet -p dora-cli -- validate examples/rust-dataflow/dataflow.yml
           cargo run --quiet -p dora-cli -- validate examples/multiple-daemons/dataflow.yml
+      # Semantic: strict-types must accept a well-typed dataflow and reject
+      # a mismatched one. Protects against the validator silently passing
+      # both cases (e.g. strict mode accidentally becoming a no-op).
+      - name: CLI semantic — dora validate --strict-types
+        shell: bash
+        run: |
+          set -e
+          cargo run --quiet -p dora-cli -- validate --strict-types examples/typed-dataflow/dataflow.yml
+          # Negative: must exit non-zero on intentional type mismatch
+          if cargo run --quiet -p dora-cli -- validate --strict-types tests/fixtures/type-mismatch.yml; then
+            echo "ERROR: --strict-types accepted tests/fixtures/type-mismatch.yml — validator regression"
+            exit 1
+          fi
       - name: CLI smoke — dora expand
         shell: bash
         run: |
           cargo run --quiet -p dora-cli -- expand examples/module-dataflow/dataflow.yml > /dev/null
+      # Semantic: expand must flatten module references into named nodes
+      # (pipeline.doubler, pipeline.filter come from the pipeline module).
+      # If expand regresses and emits the `modules:` block without inlining,
+      # this catches it.
+      - name: CLI semantic — dora expand flattens modules
+        shell: bash
+        run: |
+          set -e
+          out=$(cargo run --quiet -p dora-cli -- expand examples/module-dataflow/dataflow.yml)
+          for id in sender pipeline.doubler pipeline.filter receiver; do
+            echo "$out" | grep -q "^- id: $id" \
+              || { echo "ERROR: expand output missing node id '$id'"; exit 1; }
+          done
+          # The flattened output must not still contain a top-level modules: block
+          if echo "$out" | grep -q '^modules:'; then
+            echo "ERROR: expand output still contains 'modules:' block — not flattened"
+            exit 1
+          fi
       - name: CLI smoke — dora graph (mermaid)
         shell: bash
         run: |
           cargo run --quiet -p dora-cli -- graph examples/rust-dataflow/dataflow.yml --mermaid > /dev/null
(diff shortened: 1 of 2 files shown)


heyong4725:
No findings.

I reviewed the new semantic Tier 0 CLI assertions in `.github/workflows/ci.yml` and the new negative fixture in `tests/fixtures/type-mismatch.yml`. The added checks are exercising real command semantics rather than just invocation success, and the strict-types negative case fails for the intended reason.

I also spot-checked the exact commands locally:
- `cargo run --quiet -p dora-cli -- validate --strict-types examples/typed-dataflow/dataflow.yml`
- `cargo run --quiet -p dora-cli -- validate --strict-types tests/fixtures/type-mismatch.yml`
- `cargo run --quiet -p dora-cli -- expand examples/module-dataflow/dataflow.yml`
- `cargo run --quiet -p dora-cli -- graph examples/rust-dataflow/dataflow.yml --mermaid`

All behaved as the workflow expects.

Residual risk: the `expand` and `graph` assertions are still intentionally light semantic checks rather than golden-output comparisons, so format-level regressions could still slip through if the key tokens remain present. For Tier 0, that looks like a reasonable tradeoff.

PR #232, opened 2026-04-14
ci(nightly): add cluster lifecycle smoke job

### LINKED ISSUE #228: qa: cluster/distributed command coverage (cluster up/status/down)

heyong4725:
## Summary

Closes #228. Adds a Tier 1 nightly job covering the cluster lifecycle commands — directly addresses the "0% cluster coverage" finding in #215.

## What the job does

```bash
dora up                              # local coordinator + daemon
dora cluster status                  # assert header + >=1 daemon + "Active dataflows:"
dora run examples/rust-dataflow/dataflow.yml --stop-after 10s
dora cluster down                    # tear down
dora cluster status                  # MUST fail (coordinator gone)
pgrep -f 'dora-daemon|dora-coordinator' || true  # MUST show nothing
```

Verified locally that this sequence works end-to-end.

## What's intentionally NOT in this PR

`dora cluster up` needs SSH to each machine in a `cluster.yml`. GHA runners don't have passwordless SSH to localhost set up by default, and real remote machines are out of scope. Documented as Tier 2 (laptop/manual) in `docs/testing-matrix.md`.

## Scope

- Covers the **wire protocol** between `dora cluster status/down` and the coordinator — catches regressions where the command silently stops reporting daemons, fails to tear down, or leaks processes.
- Does **not** cover multi-machine clustering itself — that requires dedicated infra.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/nightly.yml'))"` — YAML OK
- [x] Verified locally: `dora up` → `dora cluster status` → `dora cluster down` → post-down status fails with "cannot connect to coordinator"
- [x] `pgrep -f dora-daemon` returns empty after cluster down
- [x] Updated `file-issue-on-failure` `needs:` list to include new job
- [x] Updated testing matrix Tier 1 table + Tier 2 cluster-mode section

After merge, I'll trigger the nightly via `workflow_dispatch` to confirm the new job passes.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -237,10 +237,76 @@ jobs:
       - name: Replay
         run: dora replay run.adorec
 
+  cluster-smoke:
+    # Smoke-tests the cluster lifecycle commands that don't require SSH:
+    # `dora cluster status` and `dora cluster down`. (`dora cluster up`
+    # needs SSH to each machine in cluster.yml — out of scope for a
+    # single-runner smoke; needs dedicated infra.)
+    #
+    # Strategy: use `dora up` to launch a local coordinator+daemon, then
+    # exercise the cluster CLI against it. Run a real dataflow to prove
+    # the cluster can execute work. Tear down with `dora cluster down`.
+    name: Cluster lifecycle smoke
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Build rust-dataflow example nodes
+        run: >
+          cargo build
+          -p rust-dataflow-example-node
+          -p rust-dataflow-example-status-node
+          -p rust-dataflow-example-sink
+      - name: Start local cluster (coordinator + daemon)
+        run: |
+          dora up
+          sleep 2
+      - name: cluster status shows at least 1 daemon
+        run: |
+          out=$(dora cluster status)
+          echo "$out"
+          echo "$out" | grep -q 'DAEMON ID' || { echo "ERROR: status output missing header"; exit 1; }
+          # At least one non-header line = at least one daemon row
+          daemon_lines=$(echo "$out" | grep -E '^[0-9a-f]{8}-' | wc -l)
+          test "$daemon_lines" -ge 1 || { echo "ERROR: no daemons listed"; exit 1; }
+          echo "$out" | grep -q 'Active dataflows:' || { echo "ERROR: missing dataflow summary"; exit 1; }
+      - name: Run a dataflow through the cluster
+        run: dora run examples/rust-dataflow/dataflow.yml --stop-after 10s
+      - name: cluster down tears everything down cleanly
+        run: |
+          dora cluster down
+          sleep 1
+          # Verify coordinator is actually gone: cluster status must fail.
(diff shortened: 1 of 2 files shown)


heyong4725:
1. `.github/workflows/nightly.yml:300` has a false-negative cleanup check. The job starts coordinator and daemon via `dora up`, and `binaries/cli/src/command/up.rs` launches them by re-execing the same `dora` binary with `coordinator` / `daemon` subcommands, not as `dora-coordinator` or `dora-daemon` executables. That means `pgrep -f 'dora-daemon|dora-coordinator'` will miss the actual leaked processes (`.../dora coordinator --quiet`, `.../dora daemon --quiet`) and can report clean shutdown even when `cluster down` leaves them behind. The pattern should match the real command lines, e.g. `dora (daemon|coordinator)` or otherwise inspect the process tree more directly.

I did not find a second concrete regression beyond that. I verified the workflow change against the current CLI implementation in `binaries/cli/src/command/up.rs`, `binaries/cli/src/command/cluster/status.rs`, and `binaries/cli/src/command/cluster/down.rs`, and re-validated that the YAML parses cleanly.

heyong4725:
Good catch — pushed 68c6b592.

Verified the bug locally before/after fix:

```
$ ps aux | grep -E '\bdora '
.../dora daemon --quiet
.../dora coordinator --quiet

$ pgrep -fa 'dora-daemon|dora-coordinator'    # OLD pattern
(no match — false negative)

$ pgrep -fa 'dora (daemon|coordinator)'       # NEW pattern
95344 .../dora coordinator --quiet
95345 .../dora daemon --quiet
```

Old pattern would silently let leaked processes through. New pattern matches the real subcommand invocations.

heyong4725:
No findings.

I re-checked the current PR head directly. The fix is present in `.github/workflows/nightly.yml`: the cleanup check now matches the real `dora up` process model with `pgrep -fa 'dora (daemon|coordinator)'`, which is consistent with how `binaries/cli/src/command/up.rs` launches the subprocesses.

The earlier review point is resolved. Residual risk is low and mostly operational: the check is still Linux/process-table specific, but that is fine here because `cluster-smoke` is an `ubuntu-latest` nightly job.

Issue #233, opened 2026-04-14
cli: add --count/--duration to topic echo and topic hz


heyong4725:
**Parent:** #227, #215

`dora topic echo` and `dora topic hz` stream output indefinitely with no built-in exit condition. Other inspection commands have non-interactive modes — these two don't.

## Current state

| Command | Has bounded mode? |
|---|---|
| `dora top` | `--once` ✅ |
| `dora topic list` | always exits ✅ |
| `dora topic info` | `--duration N` ✅ |
| `dora topic pub` | `--count N` ✅ |
| `dora topic echo` | streams forever ❌ |
| `dora topic hz` | streams forever (`--window N` is sliding window, not stop) ❌ |

## Why this matters

Without bounded modes, these can't be smoke-tested in CI without external `timeout` wrappers (which exit 124, not 0). They're also unfriendly for scripting — you'd have to pipe through `head -n` or similar shell tricks.

## Proposal

Add `--count <N>` and `--duration <SECONDS>` flags to both commands (whichever fires first wins, default unbounded for backward compat).

```bash
dora topic echo -d df sensor/data --count 10           # exit after 10 messages
dora topic echo -d df sensor/data --duration 5s        # exit after 5 seconds
dora topic hz -d df sensor/data --duration 5s          # exit after 5s, print summary
```

## Acceptance

- Both commands accept `--count` and `--duration` flags
- When either fires, the command exits 0 cleanly
- Default behavior unchanged (streams forever) — backward compat
- Once shipped, add nightly smoke coverage (extends `topic-and-top-smoke` job in `nightly.yml`)

## Relation to #227

`#227` is the umbrella TUI coverage issue. After this lands, all 6 commands will have CI coverage and #227 can close.

heyong4725:
Blocker on CI smoke identified and filed as #236: `_unstable_debug.publish_all_messages_to_zenoh` is not actually routing messages through Zenoh despite the flag being set.

Repro against a known-working dataflow (not specific to my fixture):
```bash
dora start examples/python-dataflow/dataflow.yml --debug --detach
dora topic info -d <name> sender/message --duration 5
# Total messages: 0
```

The CLI flag work in PR #235 is independently useful (unbounded streams are unusable in CI regardless of what the daemon does). Once #236 is fixed, we add the smoke tests here.

PR #234, opened 2026-04-14
ci(nightly): add topic + top inspection smoke job

### LINKED ISSUE #227: qa: interactive/TUI CLI coverage (topic echo/hz/info/pub, top)

heyong4725:
## Summary

Investigation phase + first wave of TUI command coverage for #227.

## Audit findings

Better than expected — most TUI commands already have non-interactive modes:

| Command | Bounded mode | This PR |
|---|---|---|
| `dora top` | `--once` (JSON snapshot) | ✅ tested |
| `dora topic list` | always exits, `--format json` | ✅ tested |
| `dora topic info` | `--duration N` | ✅ tested |
| `dora topic pub` | `--count N` | ✅ tested |
| `dora topic echo` | streams forever | ❌ filed #233 (needs `--count`/`--duration`) |
| `dora topic hz` | streams forever (`--window` is sliding, not stop) | ❌ filed #233 |

## What's added

New nightly job `topic-and-top-smoke` that exercises the 4 testable commands. Plus a fixture `tests/fixtures/zenoh-debug-dataflow.yml` (rust-dataflow with `publish_all_messages_to_zenoh: true` enabled — required for `topic info/pub` to inspect/inject runtime messages).

### Assertions

- `dora top --once` with no dataflow → must produce valid JSON array
- `dora topic list -d <df> --format json` → parsed via python, must contain `rust-node`
- `dora topic info -d <df> rust-node/random --duration 3` → output must mention the topic
- `dora topic pub -d <df> rust-node/random '42' --count 3` → exit 0, no error keywords in output

## Closes / refs

- Closes #227 for the 4 commands with bounded modes
- New issue #233 tracks the remaining 2 (`topic echo`, `topic hz`) which need new CLI flags

## Test plan

- [x] Verified `dora top --once` works (returns `[]` exit 0 with no dataflow)
- [x] Verified `dora topic list/info/pub --help` show the flags I'm using
- [x] `dora validate tests/fixtures/zenoh-debug-dataflow.yml` passes
- [x] YAML lints clean
- [x] `file-issue-on-failure` `needs:` updated to include new job
- [x] testing-matrix.md updated with new row

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -307,10 +307,73 @@ jobs:
             exit 1
           fi
 
+  topic-and-top-smoke:
+    # Smoke-tests the inspection commands: `dora top --once`, `dora topic
+    # list/info/pub`. Each has a non-interactive mode (--once / built-in
+    # exit / --duration / --count). `topic echo` and `topic hz` stream
+    # without bound — tracked separately, need --duration/--count flags.
+    name: Topic & top inspection smoke
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Build rust-dataflow example nodes
+        run: cargo build -p rust-dataflow-example-node -p rust-dataflow-example-sink
+      - name: Start coordinator + daemon
+        run: |
+          dora up
+          sleep 2
+      - name: dora top --once with no dataflow returns valid JSON
+        run: |
+          # No dataflow running yet: must produce '[]' (valid empty JSON).
+          out=$(dora top --once || true)
+          echo "$out"
+          echo "$out" | grep -qE '^\[' || { echo "ERROR: top --once didn't produce JSON array"; exit 1; }
+      - name: Start a debug-enabled dataflow
+        run: |
+          dora start tests/fixtures/zenoh-debug-dataflow.yml --name tui-smoke --detach
+          sleep 5  # let nodes register and emit at least a few messages
+      - name: dora topic list (json)
+        run: |
+          out=$(dora topic list -d tui-smoke --format json)
+          echo "$out"
+          # JSON output should be parseable and contain known node/output IDs
+          echo "$out" | python3 -c "import json,sys; data=json.load(sys.stdin); assert any('rust-node' in str(t) for t in data), f'rust-node not in {data}'"
+      - name: dora topic info on rust-node/random (--duration 3)
+        run: |
+          dora topic info -d tui-smoke rust-node/random --duration 3 2>&1 | tee info.out
+          # Should produce stats output that mentions the topic
+          grep -q 'rust-node/random' info.out || { echo "ERROR: info output missing topic"; exit 1; }
+      - name: dora topic pub --count 3
+        run: |
+          dora topic pub -d tui-smoke rust-node/random '42' --count 3 2>&1 | tee pub.out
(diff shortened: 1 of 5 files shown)


heyong4725:
1. `.github/workflows/nightly.yml:354` has a broken `topic list` assertion. `binaries/cli/src/command/topic/list.rs` does not emit a JSON array for `--format json`; it prints one JSON object per line. For the current fixture there is only one topic, so `python3 -c "import json,sys; data=json.load(sys.stdin); assert any('rust-node' in str(t) for t in data)"` parses a single object and then iterates its keys (`node`, `name`, `subscribers`), not the topic values. That assertion will fail even when the command works. The check needs to either parse NDJSON line-by-line or assert on the object fields directly.

2. `docs/testing-matrix.md:116-117` still says `dora topic echo/hz/info/pub` and `dora top` are “not covered anywhere yet,” but this PR adds nightly coverage for `top --once` and `topic list/info/pub`. That leaves the testing matrix internally inconsistent and overstates the remaining gap. The “not covered” section should be narrowed to the actually unbounded commands (`topic echo`, `topic hz`) and the still-uncovered interactive/TUI paths.

I did not find another concrete regression beyond those two from source review.

heyong4725:
Both findings addressed in 6decd7ab.

**1. NDJSON parse fix.** Verified locally: `dora topic list --format json` outputs:
```
{"node":"ticker","name":"count","subscribers":["sink/count"]}
```
Single object per line, not a JSON array. New parse:
```python
rows = [json.loads(line) for line in sys.stdin if line.strip()]
assert any(r.get('node') == 'ticker' and r.get('name') == 'count' for r in rows)
```

**2. Testing matrix.** Removed `topic info/pub/list` and `top` from the 'NOT covered' section since this PR covers them. Added an 'Already covered' subsection so the doc matches reality. Only `topic echo` and `topic hz` remain in the gap section (tracked in #233).

**Bonus fix:** the rust-dataflow fixture I started with hit its tick limit and exited before the test could query topics. Switched to Python source/sink (ticker.py / sink.py) that run forever. Required adding Python venv setup with local dora-rs to the job (noted in the PR description / commit message).

Verified end-to-end locally:
- topic list → NDJSON parse passes
- topic info → output contains `ticker/count`
- topic pub → emits `Published message 3/3`

heyong4725:
No findings.

I re-reviewed the current PR 234 head. The earlier issues are fixed:
- the `topic list --format json` check now correctly parses NDJSON line-by-line and asserts on actual fields
- `docs/testing-matrix.md` now narrows the uncovered inspection gap to `topic echo` / `topic hz` and adds an “Already covered” subsection for the bounded commands

I also checked the updated fixture path directly: the job now uses a Python ticker/sink pair so the dataflow stays alive long enough for `topic list/info/pub`, and the nightly job’s assertions are materially stronger than the previous version.

Residual risk: this still depends on the Linux nightly environment having the expected `uv` + Python behavior, but that is consistent with the job’s explicit setup and scope.

PR #235, opened 2026-04-14
cli(topic): add --count/--duration to echo, --duration to hz

### LINKED ISSUE #233: cli: add --count/--duration to topic echo and topic hz

heyong4725:
## Summary

Closes #233 (CLI changes portion). Adds bounded-execution flags to `dora topic echo` and `dora topic hz`, which previously streamed forever with no built-in exit condition.

### `dora topic echo`

```
--count <N>        Exit after this many messages (default: stream until interrupted)
--duration <SECONDS>  Exit after this many seconds (default: stream until interrupted)
```

Either fires first. Default behavior unchanged (stream forever, backward compat).

### `dora topic hz`

```
--duration <SECONDS>  Run for this many seconds without TUI, print final stats, and exit.
                      Required when stdout is not a terminal (e.g. CI, scripting).
```

When `--duration` is set, skips the ratatui TUI entirely. Collects timestamps for the given duration, then prints a plain tab-separated stats table:

```
topic               avg_ms   avg_hz   min_ms   max_ms   std_ms   samples
ALL                 100.12   9.99     99.45    101.23   0.45     30
ticker/count        100.12   9.99     99.45    101.23   0.45     30
```

When stdout is not a TTY and `--duration` is missing, the error message now points users to `--duration` (was a generic "requires interactive terminal" before).

## Why no CI smoke in this PR

Initial attempt to add a nightly smoke test exposed a separate issue: a Python-source fixture with `_unstable_debug.publish_all_messages_to_zenoh: true` doesn't actually push messages through Zenoh the way the subscribe path expects. `dora topic info` on a Running dataflow showed `Total messages: 0` after 3 seconds of collection.

This is orthogonal to the CLI flag additions. The flags are independently useful for any scripting use case right now; the smoke test will land in a follow-up once the Zenoh publish path is understood.

## Test plan

- [x] `cargo build -p dora-cli` — clean
- [x] `cargo clippy -p dora-cli -- -D warnings` — clean
- [x] `cargo fmt --all` — clean
- [x] `cargo test -p dora-cli` — passes
- [x] `dora topic echo --help` shows `--count` and `--duration`
- [x] `dora topic hz --help` shows `--duration`
- [x] Existing `hz.rs` `parse_window` tests still pass

### DIFF
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -49,12 +49,14 @@ pub struct Echo {
     #[clap(long, value_name = "FORMAT", default_value_t = OutputFormat::Table)]
     pub format: OutputFormat,
 
-    /// Exit after this many messages (default: stream until interrupted)
-    #[clap(long, value_name = "N")]
-    pub count: Option<usize>,
+    /// Exit after this many messages (default: stream until interrupted).
+    /// Must be at least 1.
+    #[clap(long, value_name = "N", value_parser = clap::value_parser!(u64).range(1..))]
+    pub count: Option<u64>,
 
-    /// Exit after this many seconds (default: stream until interrupted)
-    #[clap(long, value_name = "SECONDS")]
+    /// Exit after this many seconds (default: stream until interrupted).
+    /// Must be at least 1.
+    #[clap(long, value_name = "SECONDS", value_parser = clap::value_parser!(u64).range(1..))]
     pub duration: Option<u64>,
 
     #[clap(flatten)]
@@ -79,7 +81,7 @@ fn inspect(
     coordinator: CoordinatorOptions,
     selector: TopicSelector,
     format: OutputFormat,
-    count: Option<usize>,
+    count: Option<u64>,
     duration: Option<u64>,
 ) -> eyre::Result<()> {
     let session = coordinator.connect()?;
@@ -96,7 +98,7 @@ fn inspect(
     const HINT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
     let mut hint_shown = false;
     let mut buf = Vec::with_capacity(1024);
-    let mut emitted: usize = 0;
+    let mut emitted: u64 = 0;
     let deadline = duration.map(|s| std::time::Instant::now() + std::time::Duration::from_secs(s));
     loop {
         // Stop conditions: --count reached or --duration elapsed.
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -49,6 +49,14 @@ pub struct Echo {
     #[clap(long, value_name = "FORMAT", default_value_t = OutputFormat::Table)]
     pub format: OutputFormat,
 
+    /// Exit after this many messages (default: stream until interrupted)
+    #[clap(long, value_name = "N")]
+    pub count: Option<usize>,
+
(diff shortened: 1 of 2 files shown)


heyong4725:
1. `binaries/cli/src/command/topic/echo.rs:52-58` and `binaries/cli/src/command/topic/hz.rs:68-71` accept zero-valued bounded-mode flags, which makes the new options silently no-op instead of enforcing the “bounded execution” contract. `--count 0` on `topic echo` exits immediately without subscribing meaningfully, and `--duration 0` on either command returns immediately with empty output/stats. That is a bad failure mode for scripting/CI because a typo like `--duration 0` looks like success. These new flags should be range-checked with Clap the same way `--window` already is, e.g. `range(1..)`.

I did not find a second concrete regression beyond that from source review.

heyong4725:
Fixed in b999f818. Added `value_parser = clap::value_parser!(u64).range(1..)` to all three new flags.

Verified locally — all three now fail at parse time with exit code 2:

```
$ dora topic echo --count 0
error: invalid value '0' for '--count <N>': 0 is not in 1..18446744073709551615
exit: 2

$ dora topic echo --duration 0
error: invalid value '0' for '--duration <SECONDS>': 0 is not in 1..18446744073709551615
exit: 2

$ dora topic hz --duration 0
error: invalid value '0' for '--duration <SECONDS>': 0 is not in 1..18446744073709551615
exit: 2
```

Matches the existing `parse_window` pattern (also rejects 0). Also bumped `echo --count` from `usize` to `u64` for platform-consistent range validation.

heyong4725:
No findings.

I re-checked the latest PR 235 head directly at the new commit. The earlier issue is fixed: `topic echo --count`, `topic echo --duration`, and `topic hz --duration` now all use Clap range validation and reject `0` at parse time instead of silently succeeding with empty output.

I did not find another concrete regression in the updated bounded-mode behavior from source review.

PR #237, opened 2026-04-14
docs: clarify platform parity for C/C++ templates (#230)

### LINKED ISSUE #230: qa: platform parity gaps (C/C++ templates Linux-only)
**Parent:** #215

C and C++ template tests are Linux-only (`if: runner.os == 'Linux'` gates in `ci.yml`), even though README claims first-class macOS ARM64 and best-effort Windows support.

## The gate reasoning

The `Install Arrow C++ Library` step in the Examples job does install Arrow on macOS too (via homebrew). The gate exists because:
- Linux: `sudo apt-get install libarrow-dev` → works
- macOS: `brew install apache-arrow` → works, but header paths may differ
- Windows: no equivalent, would need vcpkg or manual CMake

### Option 1: Lift the gate for macOS

Should be mostly free — the Arrow install step already handles it. Risk: homebrew Arrow version may not match what the examples expect, flaky on macOS runner shutdowns.

### Option 2: Lift for macOS + Windows via vcpkg

Significant work. Windows doesn't have a simple package manager for C++ Arrow. Could use vcpkg-action but that's a 30-min dependency install per run.

### Option 3: Explicitly scope down

Document in `docs/testing-matrix.md`:
- C/C++ templates tested on Linux only
- macOS/Windows are best-effort, not CI-gated
- Regressions on macOS/Windows caught by users, not CI

Update README claim to match (remove "first-class macOS ARM64" for C/C++).

## Recommendation

**Option 3.** Rationale:
- The audience that uses C/C++ nodes in production is Linux-heavy
- macOS/Windows coverage for Rust/Python examples already exists
- Adding cross-platform C/C++ CI for little real-world benefit is a bad ratio

If someone reports a macOS C++ template regression, we upgrade to Option 1.

## Acceptance

- Doc updates to `README.md` and `docs/testing-matrix.md` documenting the scope
- OR (if we go Option 1): macOS `if: runner.os != 'Windows'` gate, verify examples pass on macOS CI
- `dora new --lang c/cxx` still exists and works on all platforms (already covered by `dora new` smoke in Tier 0)

heyong4725:
## Summary

Closes #230. Doc-only change. README claimed first-class macOS/ARM64 and best-effort Windows across the board, but C/C++ template tests (`dora new --lang c/cxx` + CMake build) have always been gated to `if: runner.os == 'Linux'` in CI. Document reality instead of aspiration.

## Changes

- **README.md** — Platform support table now splits Rust/Python (cross-platform) from C/C++ templates (Linux-only CI). Footnote explains where `dora new` scaffolding is smoke-tested on all platforms vs. where the full CMake build is only exercised on Linux.
- **docs/testing-matrix.md** — New \"Platform parity\" section enumerating which C/C++ tests run on which OS, with rationale for scoping (production C/C++ users are Linux-heavy; cost/benefit of maintaining Windows/macOS vcpkg setup is poor).

## Why Option 3 (document scope) over lifting the gate

Three options were considered in #230:

1. Lift the Linux gate for macOS (partially free — `Install Arrow C++ Library` already handles brew). Risk: flaky homebrew version pinning.
2. Lift for macOS + Windows via vcpkg. ~30 min install per run, significant complexity.
3. Document reality and promote gates when users report regressions.

Going with 3. Remote CI philosophy (see CLAUDE.md) is deliberately lean — promoting gates only when they catch real bugs. If a macOS/Windows C/C++ regression is reported, we lift the gate then.

## Test plan

- [x] `cargo fmt --all -- --check` — not applicable (docs only)
- [x] Rendered README platform table — reads correctly, footnote links resolve
- [x] testing-matrix anchor \`#platform-parity\` matches heading

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,10 +419,10 @@ examples/               # Example dataflows
 | WSL (x86_64) | Best effort | Best effort (not CI-gated) |
 
 C/C++ template tests (`dora new --lang c/cxx` + CMake build) run in CI on Linux only.
-The `dora new` scaffolding itself is smoke-tested on all three platforms. macOS and
-Windows C/C++ regressions are caught by users rather than CI; report them via GitHub
-issues and we will promote the gate. See [`docs/testing-matrix.md`](docs/testing-matrix.md#platform-parity)
-for the full rationale.
+`dora new --lang rust/python` is smoke-tested on all three platforms, but the C/C++
+variants are not. macOS and Windows C/C++ regressions are caught by users rather
+than CI; report them via GitHub issues and we will promote the gate. See
+[`docs/testing-matrix.md`](docs/testing-matrix.md#platform-parity) for the full rationale.
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -411,12 +411,18 @@ examples/               # Example dataflows
 
 ### Platform support
 
-| Platform | Status |
-|----------|--------|
-| Linux (x86_64, ARM64, ARM32) | First-class |
-| macOS (ARM64) | First-class |
-| Windows (x86_64) | Best effort |
-| WSL (x86_64) | Best effort |
+| Platform | Rust / Python | C / C++ templates |
+|----------|---------------|-------------------|
+| Linux (x86_64, ARM64, ARM32) | First-class | First-class (CI-gated) |
+| macOS (ARM64) | First-class | Best effort (not CI-gated) |
+| Windows (x86_64) | Best effort | Best effort (not CI-gated) |
+| WSL (x86_64) | Best effort | Best effort (not CI-gated) |
+
+C/C++ template tests (`dora new --lang c/cxx` + CMake build) run in CI on Linux only.
+The `dora new` scaffolding itself is smoke-tested on all three platforms. macOS and
+Windows C/C++ regressions are caught by users rather than CI; report them via GitHub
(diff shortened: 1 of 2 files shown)


heyong4725:
1. `README.md:422-424` overstates the current coverage. The repo does not smoke-test C/C++ `dora new` scaffolding on all three platforms; the only `dora new --lang c` / `--lang cxx` steps are still Linux-gated in `.github/workflows/ci.yml:428-443`. As written, the new footnote implies macOS/Windows at least exercise template generation while only skipping the CMake build, but CI does not currently do that.

2. `docs/testing-matrix.md:162` has the same factual mismatch. The new “Platform parity” table says `dora new --lang c/cxx` scaffolding is Tier 0 on Linux, macOS, and Windows, but the `cli` job runs those template commands only under `if: runner.os == 'Linux'`. That should be marked Linux-only unless the workflow is changed to add non-Linux scaffolding-only steps.

I did not find a second documentation mismatch beyond those two.

heyong4725:
Good catch — both instances fixed in 5e48962b.

- README footnote no longer claims `dora new` scaffolding is tested on all three platforms for C/C++; only the Rust/Python variants are.
- `docs/testing-matrix.md` table collapses `dora new --lang c/cxx` scaffolding + CMake build + `dora run` into a single Linux-only row, since they all sit under the same `if: runner.os == 'Linux'` guard in `ci.yml:428-443`.

The `dora new --lang rust/python` row stays cross-platform — those steps are not OS-gated in the `cli` job.

PR #10, opened 2026-03-10
fix: Improve error handling, security, and documentation following Rust best practices

LeonRust:
Based on Rust best practices analysis, several improvements have been made to the Adora project:

### 1. Error Handling Improvements
- **apis/python/node/src/lib.rs**: Replaced LazyLock with OnceLock for better error handling in Tokio runtime creation
  - Before: Used `expect("Failed to create Tokio runtime")` causing panic
  - After: Returns PyRuntimeError exception, allowing Python callers to catch and handle
- **apis/python/operator/src/lib.rs**: Improved error handling for lock acquisition in DelayedCleanup
  - `get_mut()` method now returns `Result<MutexGuard, String>` instead of panicking
  - Stream implementation properly handles poisoned locks and temporary unavailability

### 2. Security Dependency Updates
- **libraries/message/Cargo.toml**: Updated chrono to specific version 0.4.42 to avoid potential security issues
- **binaries/daemon/Cargo.toml**: Similarly updated chrono version for consistency

### 3. FFI Safety Enhancements
- **apis/c/node/src/lib.rs**: Added detailed SAFETY comments for unsafe operations
  - Added pointer validity guarantees for `read_adora_event_type()` function
  - Follows `#![deny(unsafe_op_in_unsafe_fn)]` requirement

### 4. Build Script Error Handling
- **apis/c++/node/build.rs**: Replaced multiple unwrap() calls with expect() for clearer error messages
  - Improved error reporting for path parsing failures
  - Provides specific context for file operation failures

### 5. Documentation Improvements
- **libraries/core/src/lib.rs**: Added detailed documentation for `adjust_shared_library_path()` function
  - Includes parameter descriptions, return values, error cases, and examples
  - Follows Rust documentation best practices

## Impact
- Increased system stability and reliability
- Improved error reporting and debugging experience
- Follows Rust security best practices
- Enhanced API documentation completeness

## Rust Best Practices Followed
1. Avoid unnecessary panics, use Result type for error returns
2. Provide safety comments for all unsafe operations
3. Maintain dependency version consistency and security
4. Provide complete public API documentation
5. Provide clear error messages in build scripts

## Testing Recommendations
- Run existing test suite to ensure backward compatibility
- Test exception handling paths for Python bindings
- Verify functionality after chrono version update

PR #238, opened 2026-04-14
fix: stream CLI topic inspection over coordinator websocket

### LINKED ISSUE #236: daemon: publish_all_messages_to_zenoh sets up pub/sub but no messages flow
## Summary

`_unstable_debug.publish_all_messages_to_zenoh: true` is set in the dataflow descriptor but messages don't actually flow through Zenoh. `dora topic info` against a running dataflow shows 0 messages received after seconds of collection.

## Repro

Patch the known-working `python-dataflow` example to enable the debug flag:

```bash
cp examples/python-dataflow/dataflow.yml /tmp/pd.yml
cat >> /tmp/pd.yml <<'EOF'

_unstable_debug:
  publish_all_messages_to_zenoh: true
EOF
cp /tmp/pd.yml examples/python-dataflow/dataflow_debug.yml

dora up
dora start examples/python-dataflow/dataflow_debug.yml --name pd-debug --detach
sleep 4

dora list   # shows `Running, 3 nodes`
dora topic info -d pd-debug sender/message --duration 5
```

Expected: after 5s of collection, `Total messages` should be > 0 (the sender emits messages at a steady rate and the dataflow successfully delivers them to receivers, as verified by `dora run` completing).

Actual:
```
Publisher:
  * sender (frequency unknown)
Subscribers:
  * transformer/message
  * receiver/message
Statistics:
  Total messages: 0
  Bandwidth: 0.00 Mbps
```

## What's confirmed

- The dataflow is genuinely running and delivering messages (sink node receives them, dataflow completes normally without the debug flag)
- Publisher/subscriber metadata is populated correctly, so `topic info` IS establishing a Zenoh subscription
- Same behavior with Rust-node fixtures
- Same behavior with `dora start --debug` as well as `_unstable_debug` in YAML

## Impact

- `dora topic echo`, `dora topic hz`, and `dora topic info --duration` are **not actually usable** for debugging real dataflows right now
- CI smoke coverage for these commands is blocked (investigated in #227, CLI flag work shipped in #235)

## Suspected cause

The daemon's `send_out()` at `binaries/daemon/src/lib.rs:2865` guards the zenoh publish behind `remote_receivers = open_external_mappings.contains(&output_id_key) || dataflow.publish_all_messages_to_zenoh`. Both sides look wired:

- `running_dataflow.rs:262` sets `publish_all_messages_to_zenoh` from the descriptor
- Coordinator's `get_or_init_zenoh` + `declare_subscriber` path works

But somewhere between `send_out()` producing the event and the zenoh subscriber receiving it, messages are being dropped. Possibilities:

1. The lazy publisher declaration at `daemon/src/lib.rs:2918` races with the subscriber — messages published before the subscriber is declared are lost (Zenoh pub/sub is transient by default)
2. `send_to_remote_receivers` is taking a different code path
3. The SHM fast path for local receivers short-circuits before the zenoh publish

Needs a debug-log trace through `send_out` -> `send_to_remote_receivers` -> actual zenoh `publisher.put()`.

## Acceptance

- `dora topic info -d <name> <topic> --duration 5` shows `Total messages > 0` for any running dataflow with `_unstable_debug.publish_all_messages_to_zenoh: true`
- Unblocks #227 (TUI smoke coverage for echo/hz)

heyong4725:
## Summary
- replace coordinator-side Zenoh topic inspection with a daemon->coordinator->CLI websocket stream for , , and 
- keep topic publish on the existing path while removing the user-facing requirement that inspection depends on 
- fix  so background websocket streaming continues after synchronous subscribe calls return, and add a regression test for late binary-frame delivery

## Why
Fixes #236. The original failure mode was late topic inspection subscribers attaching after the daemon publisher was already active and never receiving data. This patch removes that dependency from the CLI inspection path.

## Validation
- 
- 
running 1 test
test cli_topic_subscription_receives_binary_frames_after_subscribe_returns ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 15 filtered out; finished in 0.11s
- 
running 12 tests
test command::topic::hz::tests::parse_window_one ... ok
test command::topic::hz::tests::parse_window_valid ... ok
test command::topic::hz::tests::parse_window_non_numeric ... ok
test command::topic::hz::tests::parse_window_zero ... ok
test ws_client::tests::handle_response_topic_subscribe_ack ... ok
test command::tests::parse_topic_list ... ok
test command::tests::parse_topic_echo ... ok
test command::tests::parse_topic_hz ... ok
test command::tests::parse_topic_pub ... ok
test command::tests::parse_record_with_topics ... ok
test command::tests::reject_topic_pub_no_data_or_file ... ok
test command::tests::parse_topic_pub_with_file ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 120 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
- 
running 1 test
test tests::start_topic_debug_stream_targets_source_daemon ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 27 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 22 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s
- 
running 27 tests
test log::tests::log_path_format ... ok
test log::tests::log_path_rotated_format ... ok
test fault_tolerance_tests::param_delete_fails_on_closed_channel ... ok
test fault_tolerance_tests::param_delete_delivered_to_node ... ok
test fault_tolerance_tests::param_update_fails_on_closed_channel ... ok
test running_dataflow::tests::armed_deadline_past_timeout_is_timed_out ... ok
test running_dataflow::tests::armed_deadline_within_timeout_is_not_timed_out ... ok
test running_dataflow::tests::arming_a_previously_unarmed_deadline_starts_the_clock ... ok
test running_dataflow::tests::submit_to_dropped_receiver_returns_false ... ok
test fault_tolerance_tests::param_update_delivered_to_node ... ok
test running_dataflow::tests::submit_to_live_receiver_succeeds ... ok
test running_dataflow::tests::two_independent_channels_do_not_cross_deliver ... ok
test spawn::command::tests::shlex_rejects_unmatched_quote ... ok
test spawn::command::tests::shlex_splits_quoted_args ... ok
test fault_tolerance_tests::close_input_on_already_broken_input ... ok
test fault_tolerance_tests::close_input_sends_all_inputs_closed ... ok
test fault_tolerance_tests::close_input_deferred_by_broken_inputs ... ok
test fault_tolerance_tests::break_input_sends_input_closed ... ok
test fault_tolerance_tests::close_input_removes_from_open_inputs ... ok
test fault_tolerance_tests::break_input_defers_all_inputs_closed ... ok
test fault_tolerance_tests::circuit_breaker_recovery ... ok
test fault_tolerance_tests::full_circuit_breaker_cycle ... ok
test log::tests::rotate_noop_when_no_files ... ok
test log::tests::rotate_creates_dot_1_from_current ... ok
test log::tests::rotate_shifts_existing_files ... ok
test log::tests::rotate_deletes_oldest_beyond_max ... ok
test running_dataflow::tests::unarmed_deadline_is_never_timed_out ... ok

test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
-

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1844,6 +1844,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tokio-stream",
+ "tokio-tungstenite",
  "tracing",
  "uuid",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,6 +157,7 @@ futures = { workspace = true }
 tokio-stream = "0.1.11"
 tempfile = { workspace = true }
 assert2 = "0.3.16"
+tokio-tungstenite = "0.28"
 
 [[example]]
 name = "c-dataflow"
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -32,13 +32,6 @@ use crate::{
 /// Emit JSON lines:
 ///   dora topic echo -d my-dataflow robot1/pose --format json
 ///
-/// Note: The dataflow descriptor must include the following snippet so that
-/// runtime messages can be inspected:
-///
-/// ```yaml
-/// _unstable_debug:
-///   publish_all_messages_to_zenoh: true
-/// ```
 #[derive(Debug, Args)]
 #[clap(verbatim_doc_comment)]
 pub struct Echo {
@@ -127,9 +120,7 @@ fn inspect(
                 }
                 if !hint_shown {
                     eprintln!(
-                        "{}: no topic data received. Ensure your dataflow was started with \
-                         `--debug` or add the following to your dataflow YAML:\n\n  \
-                         _unstable_debug:\n    publish_all_messages_to_zenoh: true\n",
+                        "{}: no topic data received during the wait window.",
                         "hint".yellow().bold(),
                     );
                     hint_shown = true;
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -21,6 +21,13 @@ use crate::{
 /// If no `DATA` is provided, all outputs from the selected dataflow will be
 /// echoed.
 ///
+/// Topic inspection requires debug mode on the dataflow:
+///
+/// ```yaml
+/// _unstable_debug:
+///   publish_all_messages_to_zenoh: true
+/// ```
+///
 /// Examples:
 ///
 /// Echo a single topic:
@@ -120,7 +127,7 @@ fn inspect(
                 }
                 if !hint_shown {
                     eprintln!(
-                        "{}: no topic data received during the wait window.",
+                        "{}: no topic data received during the wait window. Ensure `_unstable_debug.publish_all_messages_to_zenoh: true` is enabled on the dataflow.",
                         "hint".yellow().bold(),
                     );
                     hint_shown = true;
--- a/binaries/cli/src/command/topic/hz.rs
+++ b/binaries/cli/src/command/topic/hz.rs
@@ -40,13 +40,6 @@ use crate::{
 /// Measure all topics:
 ///   dora topic hz -d my-dataflow --window 10
 ///
-/// Note: The dataflow descriptor must include the following snippet so that
-/// runtime messages can be inspected:
-///
-/// ```yaml
-/// _unstable_debug:
-///   publish_all_messages_to_zenoh: true
-/// ```
 fn parse_window(s: &str) -> Result<usize, String> {
     let val: usize = s.parse().map_err(|e| format!("{e}"))?;
     if val == 0 {
--- a/binaries/cli/src/command/topic/hz.rs
+++ b/binaries/cli/src/command/topic/hz.rs
@@ -26,6 +26,13 @@ use crate::{
 /// (average, min, max, stddev) over a sliding window. Average frequency (Hz)
 /// is derived from the average interval.
 ///
+/// Topic inspection requires debug mode on the dataflow:
+///
+/// ```yaml
+/// _unstable_debug:
+///   publish_all_messages_to_zenoh: true
+/// ```
+///
 /// If no `DATA` is provided, all outputs from the selected dataflow will be
 /// echoed.
 ///
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -25,14 +25,6 @@ use crate::{
 /// Get info for a single topic:
 ///   dora topic info -d my-dataflow camera_node/image
 ///
-/// Note: The dataflow descriptor must include the following snippet so that
-/// runtime messages can be inspected (or messages must cross machine
-/// boundaries so they are forwarded through zenoh):
-///
-/// ```yaml
-/// _unstable_debug:
-///   publish_all_messages_to_zenoh: true
-/// ```
 #[derive(Debug, Args)]
 #[clap(verbatim_doc_comment)]
 pub struct Info {
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -20,6 +20,13 @@ use crate::{
 /// Shows topic type, publisher, subscribers, and statistics (message count,
 /// bandwidth, publishing frequency).
 ///
+/// Topic inspection requires debug mode on the dataflow:
+///
+/// ```yaml
+/// _unstable_debug:
+///   publish_all_messages_to_zenoh: true
+/// ```
+///
 /// Examples:
 ///
 /// Get info for a single topic:
--- a/binaries/cli/src/ws_client.rs
+++ b/binaries/cli/src/ws_client.rs
@@ -61,7 +61,8 @@ impl WsSession {
     /// Connect to the coordinator via WebSocket.
     ///
     /// If called from within an existing tokio runtime, uses that runtime.
-    /// Otherwise creates a new single-threaded runtime.
+    /// Otherwise creates a dedicated runtime with one worker thread so the WS
+    /// receive loop keeps running after synchronous API calls return.
     pub fn connect(addr: SocketAddr) -> eyre::Result<Self> {
         if tokio::runtime::Handle::try_current().is_ok() {
             eyre::bail!(
@@ -69,7 +70,8 @@ impl WsSession {
                  use an async-native client instead"
             );
         }
-        let rt = tokio::runtime::Builder::new_current_thread()
+        let rt = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(1)
             .enable_all()
             .build()
             .context("failed to create tokio runtime for WS session")?;
@@ -188,7 +190,7 @@ impl WsSession {
             .map_err(|_| eyre!("WS session dropped reply"))?
     }
 
-    /// Subscribe to topic data via the coordinator's Zenoh proxy.
+    /// Subscribe to topic data via the coordinator's topic inspection stream.
     ///
     /// Sends a `TopicSubscribe` request, waits for the ack, then returns
     /// a `(subscription_id, receiver)` pair. Binary WS frames with matching
--- a/binaries/coordinator/src/control.rs
+++ b/binaries/coordinator/src/control.rs
@@ -30,8 +30,17 @@ pub enum ControlEvent {
     TopicSubscribe {
         dataflow_id: Uuid,
         topics: Vec<(NodeId, DataId)>,
-        /// Sends `true` if the dataflow exists and has publish_all_messages_to_zenoh enabled.
+        sender: mpsc::Sender<Vec<u8>>,
+        done_tx: oneshot::Sender<Result<Uuid, String>>,
+    },
+    TopicCheck {
+        dataflow_id: Uuid,
+        topics: Vec<(NodeId, DataId)>,
         found_tx: oneshot::Sender<bool>,
+    },
+    TopicUnsubscribe {
+        subscription_id: Uuid,
+        done_tx: oneshot::Sender<()>,
     },
     Error(eyre::Report),
 }
--- a/binaries/coordinator/src/control.rs
+++ b/binaries/coordinator/src/control.rs
@@ -30,7 +30,7 @@ pub enum ControlEvent {
     TopicSubscribe {
         dataflow_id: Uuid,
         topics: Vec<(NodeId, DataId)>,
-        sender: mpsc::Sender<Vec<u8>>,
+        sender: mpsc::Sender<crate::topic_subscriber::TopicFrame>,
         done_tx: oneshot::Sender<Result<Uuid, String>>,
     },
     TopicCheck {
--- a/binaries/coordinator/src/events.rs
+++ b/binaries/coordinator/src/events.rs
@@ -51,7 +51,7 @@ pub enum Event {
     },
     TopicDebugData {
         dataflow_id: Uuid,
-        subscription_id: Uuid,
+        subscription_ids: Vec<Uuid>,
         payload: Vec<u8>,
     },
     DaemonStatusReport {
--- a/binaries/coordinator/src/events.rs
+++ b/binaries/coordinator/src/events.rs
@@ -49,6 +49,11 @@ pub enum Event {
         metrics: BTreeMap<NodeId, NodeMetrics>,
         network: Option<NetworkMetrics>,
     },
+    TopicDebugData {
+        dataflow_id: Uuid,
+        subscription_id: Uuid,
+        payload: Vec<u8>,
+    },
     DaemonStatusReport {
         daemon_id: DaemonId,
         running_dataflows: Vec<dora_message::daemon_to_coordinator::DataflowStatusEntry>,
@@ -83,6 +88,7 @@ impl Event {
             Event::DataflowBuildResult { .. } => "DataflowBuildResult",
             Event::DataflowSpawnResult { .. } => "DataflowSpawnResult",
             Event::NodeMetrics { .. } => "NodeMetrics",
+            Event::TopicDebugData { .. } => "TopicDebugData",
             Event::DaemonStatusReport { .. } => "DaemonStatusReport",
             Event::DaemonStateCatchUpAck { .. } => "DaemonStateCatchUpAck",
         }
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -95,24 +95,40 @@ pub(crate) async fn send_topic_frames(
         );
         return;
     }
+    let shared_payload: std::sync::Arc<[u8]> = payload.into();
+    const MAX_CONSECUTIVE_TOPIC_SEND_TIMEOUTS: usize = 100;
     for subscription_id in subscription_ids {
         if let Some(subscriber) = topic_subscribers.get_mut(&subscription_id) {
-            let mut frame = Vec::with_capacity(16 + payload.len());
-            frame.extend_from_slice(&subscription_id.into_bytes());
-            frame.extend_from_slice(&payload);
+            let frame = crate::topic_subscriber::TopicFrame {
+                subscription_id,
+                payload: shared_payload.clone(),
+            };
             let send_result =
                 tokio::time::timeout(Duration::from_millis(100), subscriber.send_frame(frame))
                     .await;
             match send_result {
-                Ok(Ok(())) => {}
+                Ok(Ok(())) => {
+                    subscriber.reset_timeout_streak();
+                }
                 Ok(Err(_)) => {
                     subscriber.close();
                 }
                 Err(_) => {
-                    tracing::warn!(
-                        %subscription_id,
-                        "timed out sending topic debug frame to CLI subscriber; keeping subscription active"
-                    );
+                    let timeouts = subscriber.record_timeout();
+                    if timeouts >= MAX_CONSECUTIVE_TOPIC_SEND_TIMEOUTS {
+                        tracing::warn!(
+                            %subscription_id,
+                            timeouts,
+                            "closing slow topic debug subscriber after repeated send timeouts"
+                        );
+                        subscriber.close();
+                    } else {
+                        tracing::warn!(
+                            %subscription_id,
+                            timeouts,
+                            "timed out sending topic debug frame to CLI subscriber; keeping subscription active"
+                        );
+                    }
                 }
             }
         }
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -5,6 +5,7 @@ use crate::{
         self, ArchivedDataflow, CachedResult, DaemonConnection, DaemonConnections, RunningBuild,
         RunningDataflow,
     },
+    topic_subscriber::TopicSubscriber,
 };
 use dora_core::{
     config::{NodeId, OperatorId},
@@ -27,6 +28,8 @@ use std::{
     time::Duration,
 };
 use uuid::Uuid;
+
+const MAX_TOPIC_DEBUG_PAYLOAD_BYTES: usize = 64 * 1024 * 1024;
 
 // Resolve the dataflow name.
 pub(crate) fn resolve_name(
@@ -77,6 +80,31 @@ pub(crate) async fn send_log_message(
         }
     }
     log_subscribers.retain(|s| !s.is_closed());
+}
+
+pub(crate) async fn send_topic_frame(
+    topic_subscribers: &mut BTreeMap<Uuid, TopicSubscriber>,
+    subscription_id: Uuid,
+    payload: Vec<u8>,
+) {
+    if payload.len() > MAX_TOPIC_DEBUG_PAYLOAD_BYTES {
+        tracing::warn!(
+            "dropping oversized topic debug payload ({} bytes) for subscription {subscription_id}",
+            payload.len()
+        );
+        return;
+    }
+    if let Some(subscriber) = topic_subscribers.get_mut(&subscription_id) {
+        let mut frame = Vec::with_capacity(16 + payload.len());
+        frame.extend_from_slice(&subscription_id.into_bytes());
+        frame.extend_from_slice(&payload);
+        let send_result =
+            tokio::time::timeout(Duration::from_millis(100), subscriber.send_frame(frame)).await;
+        if send_result.is_err() {
+            subscriber.close();
+        }
+    }
+    topic_subscribers.retain(|_, s| !s.is_closed());
 }
 
 pub(crate) fn dataflow_result(
@@ -551,6 +579,7 @@ pub(crate) async fn start_dataflow(
         stop_reply_senders: Vec::new(),
         buffered_log_messages: Vec::new(),
         log_subscribers: Vec::new(),
+        topic_subscribers: BTreeMap::new(),
         daemon_ack_sequence: daemons.iter().map(|d| (d.clone(), 0)).collect(),
         pending_spawn_results: daemons,
         created_at: state::now_millis(),
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -103,8 +103,17 @@ pub(crate) async fn send_topic_frames(
             let send_result =
                 tokio::time::timeout(Duration::from_millis(100), subscriber.send_frame(frame))
                     .await;
-            if send_result.is_err() {
-                subscriber.close();
+            match send_result {
+                Ok(Ok(())) => {}
+                Ok(Err(_)) => {
+                    subscriber.close();
+                }
+                Err(_) => {
+                    tracing::warn!(
+                        %subscription_id,
+                        "timed out sending topic debug frame to CLI subscriber; keeping subscription active"
+                    );
+                }
             }
         }
     }
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -82,26 +82,30 @@ pub(crate) async fn send_log_message(
     log_subscribers.retain(|s| !s.is_closed());
 }
 
-pub(crate) async fn send_topic_frame(
+pub(crate) async fn send_topic_frames(
     topic_subscribers: &mut BTreeMap<Uuid, TopicSubscriber>,
-    subscription_id: Uuid,
+    subscription_ids: Vec<Uuid>,
     payload: Vec<u8>,
 ) {
     if payload.len() > MAX_TOPIC_DEBUG_PAYLOAD_BYTES {
         tracing::warn!(
-            "dropping oversized topic debug payload ({} bytes) for subscription {subscription_id}",
-            payload.len()
+            "dropping oversized topic debug payload ({} bytes) for {} subscription(s)",
+            payload.len(),
+            subscription_ids.len()
         );
         return;
     }
-    if let Some(subscriber) = topic_subscribers.get_mut(&subscription_id) {
-        let mut frame = Vec::with_capacity(16 + payload.len());
-        frame.extend_from_slice(&subscription_id.into_bytes());
-        frame.extend_from_slice(&payload);
-        let send_result =
-            tokio::time::timeout(Duration::from_millis(100), subscriber.send_frame(frame)).await;
-        if send_result.is_err() {
-            subscriber.close();
+    for subscription_id in subscription_ids {
+        if let Some(subscriber) = topic_subscribers.get_mut(&subscription_id) {
+            let mut frame = Vec::with_capacity(16 + payload.len());
+            frame.extend_from_slice(&subscription_id.into_bytes());
+            frame.extend_from_slice(&payload);
+            let send_result =
+                tokio::time::timeout(Duration::from_millis(100), subscriber.send_frame(frame))
+                    .await;
+            if send_result.is_err() {
+                subscriber.close();
+            }
         }
     }
     topic_subscribers.retain(|_, s| !s.is_closed());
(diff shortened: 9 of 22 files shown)


heyong4725:
## Verdict: REQUEST CHANGES

Routing approach (daemon → coordinator → CLI over websocket, bypass zenoh) is the right call for #236 — it sidesteps the late-subscribe bug entirely. But the implementation has real problems.

### CRITICAL

1. **Head-of-line blocking on daemon data path** — `binaries/daemon/src/lib.rs:3044-3059`. `send_topic_debug_frames` calls `sender.send_event().await` on a 64-slot mpsc (`coordinator.rs:114`). A slow coordinator backs up the daemon's entire output dispatch loop. At 1 kHz with two subscribers, the channel fills in ~32 ms and the daemon stalls on its data plane.

2. **Unconditional per-output serialization** — `lib.rs:2962`. `send_topic_debug_frames` is called on every output in every dataflow regardless of whether any debug subscriber exists. The empty-map guard lives inside the function; the call site doesn't skip. Pure overhead on production flows with no active inspection.

3. **Race: daemon ack vs. coordinator subscriber insertion** — `coordinator/src/lib.rs:2244-2263`. `StartTopicDebugStream` is awaited so the daemon's `debug_topic_watchers` is populated, but the coordinator only inserts the `TopicSubscriber` into its map *after* all daemon acks. Frames emitted in that window arrive at the coordinator (line 1809) and are silently dropped because the subscriber isn't registered yet. Narrow window, worst on multi-daemon dataflows. Ironically re-introduces a late-subscribe race — the same class of bug this PR is meant to fix.

### HIGH

4. **The added test doesn't reproduce #236** — `tests/ws-cli-e2e.rs:275-295`. The mock server sends the subscribe ack, sleeps 100 ms, *then* sends the binary frame. The original bug was frames arriving *before* the client finished wiring up the receiver. This test validates happy-path frame delivery after the ack is consumed — which was never broken. It cannot distinguish a fixed implementation from an unfixed one.

5. **`publish_all_messages_to_zenoh` semantics now half-decoupled.** `TopicSubscribe` bypasses the flag entirely; `TopicPublish` still gates on it and errors with `"publish_all_messages_to_zenoh not enabled"` (`ws_control.rs:279`). Users running `dora topic echo` (now works without the flag) will be confused when `dora topic pub` on the same dataflow fails citing it. Pick one story: either the flag controls both, or it controls neither.

6. **Per-subscriber JSON envelope rebuild on the hot path** — `lib.rs:3044-3055`. The `serialized_event` payload bytes are identical across subscribers; only `subscription_id` differs. The function builds a fresh JSON string containing the full payload for each subscriber. O(N × payload_size) serialization work per output per subscriber count.

### MEDIUM

7. **`tracing::info!` on every `TopicDebugData`** — `coordinator/src/lib.rs:1814`. Will spam logs at kHz rates in any real dataflow. Should be `tracing::trace!`.

8. **Scope: 777+/220-.** New message variants in both `coordinator_to_daemon.rs` and `daemon_to_coordinator.rs`, new `topic_subscriber.rs`, new `handlers::send_topic_frame`, new `ws_client::subscribe_topics`, new `ControlEvent` variants, new test harness. This is a subsystem redesign bundled as a bug fix. Review surface is large relative to the stated change.

### Must-fix before merge

1. Guard `send_topic_debug_frames` call at the call site (`lib.rs:2962`) — skip entirely when `debug_topic_watchers` is empty, not just early-return inside the function.
2. Close the ack/insertion race by inserting the `TopicSubscriber` into the coordinator map *before* sending `StartTopicDebugStream` to daemons. Roll back on daemon NACK.
3. Replace the misleading `tests/ws-cli-e2e.rs` test with one that races the frame arrival against the ack — i.e., queues the binary frame *before* the subscribe response is consumed. Otherwise we don't actually have regression coverage for #236.
4. Pick one story for `publish_all_messages_to_zenoh` + `TopicPublish`/`TopicSubscribe` gating. Either the flag is still load-bearing everywhere, or remove it from the `TopicCheck` path in `ws_control.rs:279`.
5. Downgrade the per-event `info!` to `trace!`.

### Suggestions (not blocking)

- Avoid the per-subscriber JSON rebuild; serialize the payload once and attach the `subscription_id` as a framing header, or pre-build the binary frame `[subscription_id][payload]` once per daemon emission and clone.
- Document or widen the 64-slot `CoordinatorSender` capacity. At kHz rates with a slow coordinator this will backpressure the data plane.
- The `binary_tx` channel in `ws_control.rs:75` is also 64 slots with a 100 ms drop timeout (`handlers.rs:102`). A slow terminal drops subscribers silently. Surface this to the CLI as a visible warning rather than a silent disconnect.

### What's good

- Architectural direction is right. Using the existing control-plane websocket instead of a second zenoh session is the correct solution to #236, not a workaround.
- Decoupling inspection from `publish_all_messages_to_zenoh` is a net UX win once the error-message inconsistency is fixed.
- The `ws_client.rs` fix for background streaming after synchronous subscribe returns is a legitimate bug fix independent of #236.

heyong4725:
## Follow-up: architectural direction is right

Re-read my review and it under-credits the approach. Worth saying explicitly: **the WS-based routing is the correct fix for #236, and it's better than any of the four options I floated on the original issue.**

My earlier analysis (`#236-comment-4247196973`) framed this as a zenoh problem and offered four zenoh-shaped solutions: upgrade, document the limitation, daemon re-declare on peer-join, or switch to router-mode. All four assumed "zenoh is the data plane, keep inspection on zenoh." That framing was wrong.

CLI inspection has different requirements from inter-node data:

| | Inter-node data | CLI inspection |
|---|---|---|
| Volume | kHz-MHz | Human-driven bursts |
| Latency | Microseconds | 100ms is fine |
| Topology | Many-to-many, unknown | 1-to-few, known |
| Trust | Internal | User CLI |

Zenoh is right for column one and wrong for column two. The CLI is already WS-connected to the coordinator for control. Piggy-backing inspection on that channel is the obvious answer.

This PR's scope is **narrower** than any of my four suggestions. Router-mode (my Option 4) would have restructured every daemon-to-daemon message for distributed deployments. This PR only touches inspection traffic; node-to-node data flow is unchanged.

The must-fix items in the review stand — backpressure, the ack/insertion race, the test that doesn't reproduce #236, flag consistency. But they're all implementation issues, not architectural. Reject the implementation, keep the approach.

Credit where it's due.

heyong4725:
Addressed in 66c6c8ae.

Changes against the review:
- made daemon topic-debug forwarding nonblocking with , so a slow coordinator no longer stalls the daemon output path
- gated inter-daemon serialization at the  call site when there are no remote receivers and no debug watchers
- moved coordinator  insertion ahead of daemon acks and added rollback on daemon rejection
- added coordinator regression coverage for rollback-on-failure
- tightened the WS client regression test so the binary frame is sent immediately after the subscribe ack, without the previous artificial delay
- downgraded the per-frame coordinator log to 
- restored consistent  gating for topic inspection, and updated CLI help/hints accordingly

Validation rerun after the patch:
- cargo fmt --all -- --check
- cargo test -p dora-coordinator start_topic_debug_stream
- cargo test --test ws-cli-e2e cli_topic_subscription_receives_binary_frames_immediately_after_subscribe_ack
- cargo test -p dora-cli topic
- cargo test -p dora-daemon
- cargo run --quiet -p dora-cli -- topic info -d issue236-pr ticker/count --duration 3

heyong4725:
## Re-review of commit `66c6c8ae`

Status per item from the previous review:

| # | Item | Status |
|---|---|---|
| 1 | HOL blocking (`daemon/src/lib.rs:3063`) | ✅ ADDRESSED — `try_send_event` with drop-on-full |
| 2 | Unconditional serialization (`lib.rs:2935, 2967`) | ✅ ADDRESSED — `has_debug_watchers` check at call site |
| 3 | Ack/insertion race (`coordinator/src/lib.rs:2246-2270`) | ✅ ADDRESSED — I was wrong in the original review. Coordinator inserts at `2251`, then `send_and_receive` awaits daemon reply; daemon doesn't emit until after reply. The handshake guarantees ordering. No race exists. |
| 4 | Test doesn't reproduce late-subscribe (`ws-cli-e2e.rs:273`) | ✅ ADDRESSED — renamed honestly to `..._immediately_after_subscribe_ack`, frame sent with no sleep |
| 5 | `publish_all_messages_to_zenoh` flag asymmetry (`ws_control.rs:278`) | ❌ NOT ADDRESSED |
| 6 | Per-subscriber JSON rebuild (`daemon/src/lib.rs:3052-3062`) | ❌ NOT ADDRESSED |
| 7 | `info!` spam (`coordinator/src/lib.rs:1815`) | ✅ ADDRESSED — now `trace!` |

### Correction on item 3

My original review claimed a race between daemon ack and coordinator insertion. I had the ordering wrong. The coordinator inserts the `TopicSubscriber` *before* calling `send_and_receive` on `StartTopicDebugStream`, which awaits the daemon's reply before returning. The daemon only populates `debug_topic_watchers` during that message handler, and frames can't be emitted until the handler returns. So the coordinator's map is always populated before any frame can arrive. No race. Mea culpa — that wasn't in the diff, I just mis-read the original code. Credit to the author for preserving this correct ordering.

### Remaining items

**Item 5 — flag asymmetry.** `ws_control.rs:278` still returns:
> `dataflow {id} not found, output unavailable, or \`_unstable_debug.publish_all_messages_to_zenoh\` is not enabled`

But `TopicSubscribe` now works without the flag. Running `dora topic echo` (succeeds) followed by `dora topic pub` (fails citing the flag) on the same dataflow is a confusing UX inconsistency. Fix: either remove the flag gate from `TopicPublish` (subscribe and publish both use the WS path now), or update the error message to clarify that the flag only applies to publish.

**Item 6 — per-subscriber JSON rebuild.** `daemon/src/lib.rs:3052-3062` still constructs a fresh `serde_json::to_vec(&CoordinatorRequest { subscription_id, event: serialized_event, ... })` per subscriber inside the `for subscription_id in subscription_ids` loop. The outer wrapper and `serialized_event` bytes are identical; only 16 bytes of `subscription_id` differ per iteration. For N subscribers on a kHz topic this is N × full-payload serializations where 1 × payload + N × header-swap would do. Either lift the payload serialization out of the loop and assemble the final frame as `[header_bytes][payload_bytes]` per subscriber, or send a single multi-subscriber message containing the `subscription_ids` list and fan out on the coordinator side.

### New issues introduced by the fix

None.

### Verdict: REQUEST CHANGES (minor)

Items 1, 2, 3, 4, 7 are cleanly resolved. The remaining two (5, 6) are low-risk to fix: item 5 is a string edit or one-line removal; item 6 is a 10-line restructure of the inner loop. Worth doing before merge rather than filing follow-ups — they're in the same surface area.

heyong4725:
Follow-up in d10f6d4b.

Remaining review items addressed:
- clarified the TopicPublish error to make the debug-flag dependency publish-specific
- collapsed topic-debug fanout into one daemon->coordinator event carrying the subscription ID list, so the daemon no longer rebuilds the full JSON envelope once per subscriber

Validation rerun after this patch:
- cargo fmt --all -- --check
- cargo test -p dora-coordinator start_topic_debug_stream
- cargo test -p dora-cli topic
- cargo test -p dora-daemon
- cargo run --quiet -p dora-cli -- topic info -d issue236-pr ticker/count --duration 3

heyong4725:
## Re-review of commit `d10f6d4b` "perf: collapse topic debug fanout envelopes"

### Item 6 — per-subscriber JSON rebuild: ✅ ADDRESSED

Fanout restructured correctly. Daemon sends a single `DaemonEvent::TopicDebugData { subscription_ids: Vec<Uuid>, payload: Vec<u8> }` — one serialization at `daemon/src/lib.rs:3052-3062`, one wire message regardless of subscriber count. Coordinator's `send_topic_frames` at `handlers.rs:98-109` does the per-subscriber 16-byte UUID prefix prepend inline. Payload bytes are not re-serialized. O(N) small allocs for the prefix, O(1) serialization. Textbook fix.

### Item 5 — flag asymmetry: ⚠️ PARTIALLY ADDRESSED

Gating is now symmetric — `start_topic_debug_stream` at `coordinator/src/lib.rs:2241-2244` applies the same check to both subscribe and publish. Good.

But the error message at `ws_control.rs:278` still reads:

> `dataflow {dataflow_id} not found, output unavailable, or topic publish requires \`_unstable_debug.publish_all_messages_to_zenoh: true\``

It says "topic publish requires…" even when the caller is `TopicSubscribe`. A user who runs `dora topic echo` against a dataflow without the flag will get an error saying publish is the problem. Actively misleading. One-line fix — either drop the word "publish" or generalize to "topic inspection requires".

Also worth noting: the message still conflates three distinct failures (dataflow not found, output unavailable, flag missing) into one catch-all. Not a blocker, but a user won't know which of the three hit them.

### Nits (non-blocking)

1. **Wire protocol breaking change.** `DaemonEvent::TopicDebugData` now carries `subscription_ids: Vec<Uuid>` instead of a single `subscription_id`. Safe in practice because daemon and coordinator are always co-deployed from the same binary, but the struct at `coordinator/src/events.rs:52-56` has no version guard. Mention it in commit message or release notes.

2. **Redundant clone at `daemon/src/lib.rs:3057`** — `subscription_ids: subscription_ids.clone()` where the value is immediately moved into the struct. Should be a move, not a clone.

No regressions on items 1–4, 7.

### Verdict: APPROVE WITH NITS

Item 6 fully closed. Item 5 functionally fixed but the error string lies — please fix `ws_control.rs:278` before merge, it's a ~5-character edit. Once that's done, ship it.

heyong4725:
## Full-PR /review + /simplify pass

Re-reviewed all three commits (`dc8f8336`, `66c6c8ae`, `d10f6d4b`) as a combined delta (~1051+/225-) against `main`. Previous per-commit reviews focused on targeted fixes; this one asks whether the end state is shippable as a whole. Combined findings from correctness + simplification passes, deduped and prioritized.

### CRITICAL

1. **Dataflow finish leaks daemon watchers and silently hangs CLI subscribers.** `coordinator/src/lib.rs:483` drops `finished_dataflow.topic_subscribers` without sending `StopTopicDebugStream` to daemons or closing subscriber channels. Daemon's `debug_topic_watchers` / `debug_topic_subscriptions` leak for the daemon's lifetime. CLI's `data_rx` just stops receiving frames with no EOF — `dora topic echo` appears to freeze rather than report "dataflow finished." Fix: on `DataflowFinishedOnDaemon`, iterate `topic_subscribers`, send `StopTopicDebugStream` to each daemon, and close each subscriber's channel so the CLI sees clean EOF.

2. **`send_topic_frames` conflates timeout with channel close.** `handlers.rs:106` — `send_result.is_err()` is `true` for both `Elapsed` (100ms timeout, slow consumer) and the real `Closed` variant. On timeout the subscriber is permanently evicted with no log and no CLI-visible error. A CLI that stalls for 100ms (e.g. terminal scrollback redraw) gets silently disconnected. Fix: distinguish `Err(Elapsed)` → warn + retry, from `Ok(Err(_))` → close.

### HIGH

3. **Daemon reconnect doesn't re-register watchers.** When a daemon reconnects after transient failure, the coordinator re-associates the connection but never re-sends `StartTopicDebugStream` for active subscribers. Inspection silently goes dark. Fix: on daemon reconnect, walk `running_dataflows[*].topic_subscribers` and re-issue `StartTopicDebugStream` to the reconnected daemon.

4. **Double-subscribe produces duplicate frames.** Same CLI calling `TopicSubscribe` twice for `(dataflow_id, topics)` creates two `subscription_id`s and two daemon watchers. CLI gets 2× frames until both are explicitly unsubscribed. No idempotency check. Fix: dedupe by `(connection_id, dataflow_id, topics)` tuple, or document the non-idempotent behavior.

5. **`publish_topic` lost its zenoh session cache.** `ws_control.rs:425` — previously `zenoh_session` was lazily cached per connection; the rewrite dropped the cache because subscribe no longer needs zenoh, but `publish_topic` still does. Every `dora topic pub` call now pays a full zenoh session open. Fix: restore a per-connection `Option<zenoh::Session>` or share one coordinator-wide session for publish.

6. **`stop_topic_debug_stream` aborts teardown loop on first missing daemon.** `coordinator/src/lib.rs:2373-2395` — `get_mut(&daemon_id).wrap_err(...)?` bubbles up and stops the loop. After fix #1 lands, this will cause teardown to skip remaining daemons whenever a multi-daemon dataflow has already lost one connection. Fix: log and continue, don't propagate.

### MEDIUM

7. **Sequential daemon fanout in start/stop.** `coordinator/src/lib.rs:2259-2304` and `2378-2400` — `send_and_receive` awaited serially per daemon. For N daemons this is N round-trips. Fix: `futures::future::try_join_all`.

8. **Payload clone per subscriber in coordinator fanout.** `handlers.rs:100-101` — `Vec::with_capacity(16 + payload.len())` and full payload clone per subscriber. For K subscribers × M-byte payload it's K×M bytes copied in the coordinator hot path. Already improved from the original (daemon-side single-serialize) but the coordinator side still duplicates. Fix: `Arc<[u8]>` for the payload, or switch the WS sink to a vectored-send API (`[prefix_bytes, payload_arc]`).

9. **Redundant inverse state on daemon.** `daemon/src/running_dataflow.rs:204-205` — `debug_topic_watchers: BTreeMap<OutputId, BTreeSet<Uuid>>` and `debug_topic_subscriptions: BTreeMap<Uuid, BTreeSet<OutputId>>` are strict inverses. Two insert sites, two remove sites in `lib.rs:1820-1875`. Maintenance trap. Fix: keep only the hot-path lookup (watchers) and iterate to unsubscribe, or wrap the pair in a small struct that enforces the invariant.

10. **Flag name `publish_all_messages_to_zenoh` is now misleading.** The feature no longer publishes to zenoh for inspection. Rename to `enable_debug_inspection` (or similar), keep the old name as an alias for one release, deprecate. Future maintainers will be confused.

### LOW

11. **Dead binding.** `coordinator/src/lib.rs:2324` — `let _outputs_by_daemon = subscriber.outputs_by_daemon().clone();` is a no-op that allocates a HashMap clone for nothing. Delete.

12. **`.clone()` before move into struct.** `daemon/src/lib.rs:3057-3058` — `subscription_ids: subscription_ids.clone()` clones just to borrow `.len()` later. Fix: `let n = subscription_ids.len(); ... subscription_ids,` — move, not clone.

13. **`serialized_event.to_vec()` on hot path.** `daemon/src/lib.rs:3059` — when debug is the sole consumer (no remote receivers), the `to_vec()` is avoidable by threading ownership. Minor.

14. **Wire protocol change undocumented.** `TopicDebugData { subscription_ids: Vec<Uuid>, payload: Vec<u8> }` replaces a single-uuid variant. Safe under co-deployment but worth a note in commit message or `libraries/message/` changelog comment.

### Test coverage gaps

The new e2e test (`ws-cli-e2e.rs:273`) uses a handcrafted mock server and exercises only `ws_client::subscribe_topics` — not the real coordinator's `start_topic_debug_stream`, the daemon's `StartTopicDebugStream` handler, or `send_topic_frames`. None of the lifecycle cases are tested: CLI disconnect, dataflow finish, daemon reconnect, double-subscribe, slow-consumer eviction. At minimum, add an integration test that runs coordinator + mock daemon, starts a subscription, stops the dataflow, asserts clean EOF on `data_rx` (regression guard for #1).

### Scope note

Test infra changes (`coordinator/tests/common/mod.rs` global mutex + `TestCoordinatorHandle`, flipping 23 tests to `multi_thread`) appear necessary for the new WS client runtime, not scope creep — but worth a line in the PR description.

### What's good

- Routing approach itself is correct — reviewed and credited in prior comments.
- Rollback-on-failure in `start_topic_debug_stream` is well-handled.
- Daemon-side single-serialization (`d10f6d4b`) is the right pattern.
- `try_send_event` backpressure on the daemon output loop is the right call.
- CLI side is properly shared through `ws_client::subscribe_topics` — no copy-paste across `echo.rs`/`hz.rs`/`info.rs`.
- No unsafe, no new panics, error handling is on par with surrounding code.

### Verdict: REQUEST CHANGES

Items 1 and 2 are blockers — the dataflow-finish leak + silent CLI hang, and the timeout-eviction conflation, will both ship user-visible regressions the moment this hits a real workload. Items 3, 4, 5, 6 are all direct consequences of the new architecture and deserve to land in this PR rather than follow-up. Items 7-14 are fine as a follow-up but cheap to fold in now.

Core direction is right, commit 1→2→3 show clean incremental response to feedback. Not far from shippable.

heyong4725:
## Re-review after commits `90753d63` + `97589f45`

Status per item from the last full-PR review:

| # | Item | Status |
|---|---|---|
| 1 | Dataflow finish leak / silent CLI hang (CRITICAL) | ✅ ADDRESSED — `coordinator/src/lib.rs:484-486` iterates `topic_subscribers.values_mut()` and calls `subscriber.close()`; dropping the `mpsc::Sender` signals EOF to the CLI reader |
| 2 | Timeout = silent eviction (CRITICAL) | ⚠️ PARTIALLY ADDRESSED — see below |
| 3 | Daemon reconnect doesn't re-register (HIGH) | ✅ ADDRESSED — `restore_topic_debug_streams_for_daemon` at `lib.rs:2451-2531`, invoked on reconnect at `lib.rs:2026` |
| 4 | Double-subscribe duplicate frames (HIGH) | ❌ NOT ADDRESSED |
| 5 | `publish_topic` zenoh session cache (HIGH) | ✅ ADDRESSED — `ws_control.rs:419` now takes `&mut Option<zenoh::Session>`, lazy-init once per connection |
| 6 | `stop_topic_debug_stream` aborts on missing daemon (HIGH) | ✅ ADDRESSED — `lib.rs:2396-2402` logs and `continue`s |
| 7 | Sequential daemon fanout (MED) | ✅ ADDRESSED — `join_all` on both start (`2267-2315`) and stop (`2394-2442`). Each future clones its `DaemonConnection`; no shared mutable state. |
| 8 | Per-subscriber payload clone on coordinator (MED) | ❌ NOT ADDRESSED |
| 9 | Redundant inverse maps on daemon (MED) | ❌ NOT ADDRESSED |
| 10 | Flag name misleading (MED) | ❌ NOT ADDRESSED |
| 11 | Dead `_outputs_by_daemon` binding (LOW) | ✅ ADDRESSED — deleted |
| 12 | `subscription_ids.clone()` (LOW) | ✅ ADDRESSED — moved, not cloned |
| 13 | `serialized_event.to_vec()` (LOW) | ✅ ADDRESSED — conditional clone only when both debug + remote receivers are active |

### Detail on the partial fix (#2)

`handlers.rs:111-116` now logs a warning on timeout and keeps the subscription alive instead of evicting. The immediate correctness bug (CLI silently dropped on a 100ms stall) is gone — good.

But the new behavior has its own failure mode: an infinitely-slow CLI consumer now accumulates repeated warns while the coordinator continues to `frame.extend_from_slice(&payload)` per subscriber, per frame, forever. No backpressure escalation. The subscription never closes, the warn log never stops, and coordinator CPU/memory grow with every dropped frame. Reduced from silent correctness bug to bounded-but-uncontrolled resource use. Suggested escalation: a per-subscriber `consecutive_timeouts` counter that closes the subscription after N (say 100) consecutive failures, with a final warn that actually tells the operator what happened.

This is coupled to #8 — the fix is the same surface. Wrap `payload` in `Arc<[u8]>` once at the top of `send_topic_frames`, give each subscriber only the 16-byte UUID prefix + Arc-shared payload, and let the subscriber's own bounded mpsc provide natural backpressure. Both issues resolve in the same 20 lines.

### Still open

- **#4 double-subscribe.** `start_topic_debug_stream` at `lib.rs:2257` always calls `Uuid::new_v4()` with no dedup guard. Fix is small: check for an existing `(dataflow_id, topics)` entry before generating a new UUID, or document the non-idempotency.
- **#8 per-subscriber payload clone** — see #2 above.
- **#9 redundant inverse maps** — lower priority, cosmetic maintenance trap.
- **#10 flag rename** — deferrable, but worth a follow-up issue so future maintainers don't trip.
- **Lifecycle test gap.** Two new unit tests cover the parallel rollback path (good). No integration test for dataflow-finish EOF delivery, reconnect re-registration, double-subscribe, or slow-consumer handling. At minimum #1 (the resolved critical) deserves a regression test — otherwise a future refactor can silently reintroduce it.

### New issues

None. The parallel refactor in `97589f45` is clean. Cloned `DaemonConnection` per-future matches the existing `destroy_daemons` pattern; error collection (accumulate all, return first) is correct.

### Verdict: APPROVE WITH NITS

Both CRITICAL blockers resolved. Three of four HIGH issues resolved. The remaining items are:

- **#2 partial + #8 together** — one focused fix (Arc-share payload + escalate on consecutive timeouts). Would be nice in this PR; acceptable as an immediate follow-on.
- **#4 double-subscribe** — small, worth fixing here.
- **#9, #10, lifecycle tests** — follow-up issues are fine.

If the author addresses #4 and the combined #2/#8 fix in this PR, ship it. If they'd rather file #4 and #2/#8 as follow-ons with concrete tracking issues, I'd still approve — neither item produces an immediate user-visible failure, just bounded resource waste and potential duplicate frames.

heyong4725:
## Re-review after commit `3f640bd5`

Status per item from the last review:

| # | Item | Status |
|---|---|---|
| 2 + 8 | Payload clone + timeout escalation (coupled, HIGH) | ✅ ADDRESSED |
| 4 | Double-subscribe idempotency (HIGH) | ❌ NOT ADDRESSED — see calibration below |
| 9 | Redundant inverse maps on daemon (MED) | ❌ NOT ADDRESSED |
| 10 | Flag rename (MED) | ❌ NOT ADDRESSED |
| — | Lifecycle test gap | ❌ NOT ADDRESSED |

### Items 2 + 8 — well done

`handlers.rs:98` converts `payload` to `Arc<[u8]>` once before the subscriber loop; each iteration clones only the pointer. `topic_subscriber.rs:14` adds `consecutive_timeouts: usize` with `record_timeout()` / `reset_timeout_streak()`; `handlers.rs:118` closes the subscriber at `MAX_CONSECUTIVE_TOPIC_SEND_TIMEOUTS=100`. Both zero-copy sharing and backpressure escalation implemented cleanly in one focused commit.

One small nit: the `100` constant is declared inline inside `send_topic_frames` while `MAX_TOPIC_DEBUG_PAYLOAD_BYTES` is at module scope. Normalize both to module-level `const` for consistency.

### Calibrating item 4

Previous review labeled double-subscribe as a correctness bug. Spot-checked the actual CLI call sites:
- `binaries/cli/src/command/topic/echo.rs:95` — one `subscribe_topics` per invocation
- `binaries/cli/src/command/topic/hz.rs:88` — same
- `binaries/cli/src/command/topic/info.rs:152` — same

Each CLI subcommand opens a fresh WS connection and subscribes once. No shipped code path triggers the duplicate-frame case. This is a robustness/defensive concern, not a user-visible bug. I was too strict earlier.

Real-world trigger would be: a third-party client built on the WS control protocol that calls `TopicSubscribe` twice on the same connection. Worth fixing for API robustness, but fine as a follow-up issue.

### What's still open

All safe-to-defer as follow-up issues:

- **#4** — dedup check on `(dataflow_id, sorted_topics, connection_id)` before generating a new UUID, or document non-idempotency in the protocol docs
- **#9** — collapse inverse maps on daemon (cosmetic maintenance trap)
- **#10** — rename `publish_all_messages_to_zenoh` → `enable_debug_inspection`, keep old name as deprecated alias for one release
- **Lifecycle tests** — integration tests for dataflow-finish EOF delivery, daemon reconnect re-registration, slow-consumer 100-timeout close

Of these, the test gap is the most important — the CRITICAL fix (#1 dataflow-finish cleanup) currently has no regression coverage, and the elaborate new machinery (reconnect restore, backpressure escalation) is an easy target for silent future regressions.

### No new issues introduced

`Arc<[u8]>` conversion point is correct — done before the loop, not inside it. `record_timeout`/`reset_timeout_streak` ownership is clean. No lifetime or Send/Sync concerns with the refactor.

### Verdict: APPROVE WITH NITS

All CRITICAL and effectively all HIGH items from the full-PR review are now resolved. The PR went from "shouldn't merge" to "ready" over four iterations. Remaining items are legitimately follow-up material.

**Ship recommendation:**
1. Land this PR as-is, or fold in the one-line constant normalization nit
2. File follow-up issues for #4 (API robustness), #9 + #10 (cleanup), and integration test coverage — don't let them float as TODOs
3. Close #236 when this merges

Good iteration cycle. The incremental commits (`dc8f...` → `66c6...` → `d10f...` → `9075...` → `9758...` → `3f64...`) each addressed specific review feedback without scope drift.

heyong4725:
## Follow-up issues filed

Four follow-ups for items deferred out of this PR:

- dora-rs/adora#239 — topic debug: dedup on double-subscribe (API robustness, low priority)
- dora-rs/adora#240 — rename `publish_all_messages_to_zenoh` → `enable_debug_inspection` (medium, naming)
- dora-rs/adora#241 — collapse redundant debug topic inverse maps on daemon (medium, maintenance)
- dora-rs/adora#242 — integration tests for lifecycle cases: dataflow-finish EOF, daemon reconnect, slow-consumer close (high priority for regression safety)

#242 is the one worth prioritizing — the lifecycle code added in this PR is exactly the kind of subtle control-plane logic that silently regresses on future refactors.

Closing #236 when this merges.

PR #244, opened 2026-04-15
test(coordinator): lifecycle coverage for topic debug subscribers (#242)

### LINKED ISSUE #242: topic debug: integration tests for lifecycle cases
Follow-up from #238. High priority for regression safety.

## Problem

#238 shipped with unit tests for the parallel rollback path and the new `ws_client` binary-frame delivery, but **no integration coverage** for the lifecycle machinery added to fix review blockers:

1. **Dataflow-finish EOF** — when a dataflow stops, `coordinator/src/lib.rs:484-486` closes subscriber channels. CLI should see clean EOF on `data_rx`. This was the CRITICAL fix from review round 3. No regression test.
2. **Daemon reconnect re-registration** — `restore_topic_debug_streams_for_daemon` at `coordinator/src/lib.rs:2451-2531` is invoked on reconnect. No test that this actually re-issues `StartTopicDebugStream` correctly.
3. **Slow-consumer backpressure** — `handlers.rs:118` closes the subscription after `MAX_CONSECUTIVE_TOPIC_SEND_TIMEOUTS = 100` failures. No test that the counter increments/resets correctly, or that close fires at the threshold.

## Plan

Add integration tests under `binaries/coordinator/tests/` or `tests/ws-cli-e2e.rs` using a real coordinator + mock daemon:

- `subscribe → dataflow finish → data_rx sees EOF`
- `subscribe → daemon disconnect → daemon reconnect → subscription still delivers frames`
- `subscribe → consumer stalls → after 100 consecutive timeouts, subscriber is closed with a warn log`

## Why it matters

The reconnect and lifecycle code was added in direct response to review feedback. It's exactly the kind of subtle control-plane logic that silently regresses on future refactors. Regression tests lock in the current behavior.

## Acceptance

- At least one test per scenario above, running in Tier 0 CI

heyong4725:
## Summary

Closes #242. Integration-style unit tests locking in the three pieces of control-plane behavior added in #238. All 8 tests run in ~10ms total.

### Dataflow-finish cleanup (the CRITICAL fix from #238)

- `close_signals_eof_to_receiver` (`topic_subscriber.rs`): direct test of `TopicSubscriber::close()` — receiver sees `None`, not hang
- `send_frame_fails_after_close` (`topic_subscriber.rs`): `send_frame` errors after close
- `dataflow_finish_closes_all_topic_subscribers` (`lib.rs`): mirrors the production cleanup loop at `lib.rs:484-486` with two subscribers; asserts both see EOF

### Daemon reconnect re-registration

- `restore_topic_debug_streams_re_issues_start_after_reconnect`: pre-populates `topic_subscribers`, runs `restore_topic_debug_streams_for_daemon`, asserts the reconnected daemon receives a fresh `StartTopicDebugStream` with the correct subscription_id + dataflow_id
- `restore_topic_debug_streams_skips_unreported_dataflows`: empty `reported_dataflows` set → no message sent

### Slow-consumer backpressure escalation

- `send_topic_frames_closes_subscriber_after_100_consecutive_timeouts`: capacity-1 channel pre-filled, rx never drained, loop 100 times. Uses `#[tokio::test(flavor = "current_thread", start_paused = true)]` so the 100 × 100ms timeout sequence runs instantly without real wall-clock delay. Asserts eviction happens exactly at iteration 100.
- `send_topic_frames_resets_timeout_streak_on_successful_send`: primes a 3-timeout streak, then a real send resets the counter
- `record_timeout_is_monotonic`: unit coverage for the counter

## Why this matters

The #238 work added subtle control-plane logic: dataflow-finish cleanup, reconnect restore, backpressure escalation. Each piece is the kind of thing that silently regresses on future refactors — the behavior is observable only at the CLI/UX layer, which is exactly where regressions go unnoticed in CI.

These tests exercise each primitive directly against the real production code paths, not mocks. A future change that e.g. stops calling `close()` in the dataflow-finish handler will fail `dataflow_finish_closes_all_topic_subscribers`. A change that breaks the reconnect restore flow fails `restore_topic_debug_streams_re_issues_start_after_reconnect`.

## Test plan

- [x] `cargo test -p dora-coordinator` — all 8 new + existing tests pass
- [x] Test runtime: all 8 tests complete in ~10ms (100-timeout test runs in ~0ms via paused time)

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -667,3 +667,70 @@ async fn destroy_daemons(
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::topic_subscriber::TopicSubscriber;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn send_topic_frames_resets_timeout_streak_on_successful_send() {
+        let subscription_id = Uuid::new_v4();
+        let (tx, mut rx) = tokio::sync::mpsc::channel(4);
+        let mut subscriber = TopicSubscriber::new(BTreeMap::new(), tx);
+        // Prime the counter: simulate 3 prior timeouts.
+        assert_eq!(subscriber.record_timeout(), 1);
+        assert_eq!(subscriber.record_timeout(), 2);
+        assert_eq!(subscriber.record_timeout(), 3);
+
+        let mut subscribers = BTreeMap::new();
+        subscribers.insert(subscription_id, subscriber);
+
+        // A real send should succeed (channel has capacity) and reset the streak.
+        send_topic_frames(&mut subscribers, vec![subscription_id], vec![1, 2, 3]).await;
+
+        let frame = rx.try_recv().expect("subscriber should receive frame");
+        assert_eq!(&*frame.payload, &[1u8, 2, 3]);
+        // Streak must be back to 0 after a successful send.
+        assert_eq!(
+            subscribers
+                .get_mut(&subscription_id)
+                .unwrap()
+                .record_timeout(),
+            1
+        );
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn send_topic_frames_closes_subscriber_after_100_consecutive_timeouts() {
+        // Capacity 1 + pre-filled + rx never drained = every send_frame times out.
+        // `start_paused = true` makes tokio auto-advance through the 100 × 100ms
+        // timeouts without real wall-clock delay.
+        let subscription_id = Uuid::new_v4();
+        let (tx, _rx_never_drained) = tokio::sync::mpsc::channel(1);
+        tx.send(crate::topic_subscriber::TopicFrame {
+            subscription_id,
+            payload: std::sync::Arc::from(vec![].into_boxed_slice()),
+        })
+        .await
+        .unwrap();
+
+        let subscriber = TopicSubscriber::new(BTreeMap::new(), tx);
+        let mut subscribers = BTreeMap::new();
+        subscribers.insert(subscription_id, subscriber);
+
+        for i in 1..=99 {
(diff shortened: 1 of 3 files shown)


heyong4725:
No findings.

I reviewed the new coordinator lifecycle tests in binaries/coordinator/src/handlers.rs, binaries/coordinator/src/lib.rs, and binaries/coordinator/src/topic_subscriber.rs against the production paths they are meant to lock in. The added coverage is narrow but useful: it exercises the slow-consumer timeout threshold, reconnect re-registration, and the dataflow-finish close behavior that mattered in #238, without broad mocking or brittle timing.

Validation:
- cargo test -p dora-coordinator

Residual risk: dataflow_finish_closes_all_topic_subscribers still mirrors the production cleanup loop rather than invoking the full event-loop path end-to-end, so it protects the close semantics but not the higher-level dispatch wiring around that branch.

heyong4725:
Addressing the residual-risk note in 112f9f8e. The full event-loop harness turned out to need ~150 lines of mock-daemon scaffolding (spawn/build replies) that didn't feel proportionate. Took a middle path:

1. Extracted the inline cleanup into `close_topic_subscribers_on_finish` — now a named function with exactly one caller. If someone removes the call site, the helper becomes dead code and trips `dead_code` lint.
2. Bound the existing test to the real helper instead of a mirror (`super::close_topic_subscribers_on_finish`). Changes to helper semantics fail the test directly.
3. Added `dataflow_finish_dispatch_calls_close_helper` — source-level guard using `include_str!("lib.rs")` that asserts the event-loop arm still calls the helper. Not elegant, but it hard-fails the specific regression you flagged (someone moves the branch, forgets to keep cleanup wired up). The test message tells future maintainers exactly what to update if they intentionally relocate the call.

Two layers of protection: dead_code lint catches removal, source-grep catches silent breakage of the wiring. Short of a full mock-daemon harness, this is as tight as I can make it without disproportionate scaffolding.

PR #245, opened 2026-04-15
refactor: rename publish_all_messages_to_zenoh → enable_debug_inspection (#240)

### LINKED ISSUE #240: rename: publish_all_messages_to_zenoh → enable_debug_inspection
Follow-up from #238. Medium priority, naming/UX.

## Problem

`_unstable_debug.publish_all_messages_to_zenoh` in the dataflow descriptor now gates CLI topic inspection over the coordinator WebSocket, not zenoh. The flag name lies.

Touched in: `dora-schema.json`, `binaries/daemon/src/running_dataflow.rs:210`, `binaries/coordinator/src/ws_control.rs`, docs.

## Fix

1. Rename field to `enable_debug_inspection` (or similar)
2. Keep `publish_all_messages_to_zenoh` as a deprecated alias that maps to the new field with a warn-on-load for one release
3. Update schema, docs, and error messages (see `ws_control.rs:278` which still references the old name in its error string)
4. Remove alias in the following release

## Why this isn't critical

The flag is behind `_unstable_debug`, so consumers have been warned. Still worth doing before the flag moves out of the unstable namespace.

## Acceptance

- Schema + field renamed with deprecation shim
- Error messages updated
- Docs + README updated

heyong4725:
## Summary

Closes #240. The flag `publish_all_messages_to_zenoh` has been misleading since #238 landed — it no longer publishes to zenoh for CLI inspection. It gates the coordinator WebSocket debug stream. Renamed to `enable_debug_inspection`, which reflects what the flag actually does.

## Backward compatibility

The serde alias on the `Debug` struct keeps existing dataflow YAML parsing:

```rust
#[serde(default, alias = "publish_all_messages_to_zenoh")]
pub enable_debug_inspection: bool,
```

Two unit tests in `libraries/message/src/descriptor.rs` lock this in:
- `debug_flag_accepts_new_name` — new spelling works
- `debug_flag_accepts_legacy_alias` — old spelling still deserializes into the new field

## What changed

- **`libraries/message/src/descriptor.rs`** — field rename + serde alias + 2 tests
- **`binaries/{coordinator,daemon,cli}/**.rs`** — every call site updated to the new field name
- **Error messages** in `ws_control.rs` and `coordinator/lib.rs` now mention both names so users hitting the error know the migration happened
- **`docs/*.md`, `guide/src/**/*.md`** — sed rename across all user-facing docs
- **`dora-schema.json`** — documents both names, with the old one flagged as deprecated
- **`tests/fixtures/zenoh-debug-dataflow.yml`** — uses the new name so CI smoke exercises the new path

## Test plan

- [x] `cargo test --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python --exclude dora-cli-api-python --exclude dora-examples` — all pass, no failures
- [x] `debug_flag_accepts_legacy_alias` test proves existing YAML files keep working

## Follow-up (out of scope)

Removal of the serde alias itself — do it in a later release once users have had time to migrate. File a tracking issue near that release.

### DIFF
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -189,7 +189,7 @@ impl Executable for Run {
             .expand(working_dir)
             .wrap_err("failed to expand modules in dataflow descriptor")?;
         if self.debug {
-            dataflow_descriptor.debug.publish_all_messages_to_zenoh = true;
+            dataflow_descriptor.debug.enable_debug_inspection = true;
         }
 
         // Validate: dora run doesn't support deploy keys
--- a/binaries/cli/src/command/start/mod.rs
+++ b/binaries/cli/src/command/start/mod.rs
@@ -131,7 +131,7 @@ fn start_dataflow(
         DataflowSession::read_session(&dataflow).context("failed to read DataflowSession")?;
 
     if debug {
-        dataflow_descriptor.debug.publish_all_messages_to_zenoh = true;
+        dataflow_descriptor.debug.enable_debug_inspection = true;
     }
 
     let session = connect_to_coordinator(coordinator_socket)?;
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -25,7 +25,7 @@ use crate::{
 ///
 /// ```yaml
 /// _unstable_debug:
-///   publish_all_messages_to_zenoh: true
+///   enable_debug_inspection: true
 /// ```
 ///
 /// Examples:
@@ -127,7 +127,7 @@ fn inspect(
                 }
                 if !hint_shown {
                     eprintln!(
-                        "{}: no topic data received during the wait window. Ensure `_unstable_debug.publish_all_messages_to_zenoh: true` is enabled on the dataflow.",
+                        "{}: no topic data received during the wait window. Ensure `_unstable_debug.enable_debug_inspection: true` is enabled on the dataflow.",
                         "hint".yellow().bold(),
                     );
                     hint_shown = true;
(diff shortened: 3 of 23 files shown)


heyong4725:
1. guide/po/messages.pot:5500 and guide/po/zh-CN.po:5771 still contain the old publish_all_messages_to_zenoh wording in multiple guide strings, even though this PR updates the English guide sources to enable_debug_inspection. That leaves the committed gettext sources out of sync with the docs and will keep localized/translation workflows surfacing the deprecated name as if it were current. Since this PR is a user-facing terminology rename, the .pot and checked-in .po artifacts should be regenerated or updated in the same change.

I did not find a runtime/code-path regression in the actual rename. Validation:
- cargo test -p dora-message debug_flag_accepts

heyong4725:
Addressed in 95fa8fbc — mechanical sed across `guide/po/messages.pot` and `guide/po/zh-CN.po` so the translation msgids match the English docs. 26 replacements per file, verified no `publish_all_messages_to_zenoh` remains in either.

Only caveat: the zh-CN translations (msgstr) still render the old term in Chinese prose until a translator revisits them. That's a separate translation task, not a rename-consistency issue — the msgids now correctly key off the new English term, so future translators will see the right source text to translate.

heyong4725:
No findings.

I re-reviewed PR 245 at its current head. The earlier doc-sync issue is fixed: guide/po/messages.pot and guide/po/zh-CN.po now track the new enable_debug_inspection name consistently, so the committed gettext artifacts match the updated English docs.

I also re-checked the compatibility path in libraries/message/src/descriptor.rs; the serde alias still preserves old YAML, and cargo test -p dora-message debug_flag_accepts passes locally.

Residual risk: this is still a terminology migration with a long-tail docs footprint, so the main remaining risk is an overlooked external doc/example outside this repo rather than a code-path regression here.

PR #246, opened 2026-04-15
refactor(daemon): collapse debug topic inverse maps (#241)

### LINKED ISSUE #241: daemon: collapse redundant debug topic inverse maps
Follow-up from #238. Medium priority, maintenance.

## Problem

`binaries/daemon/src/running_dataflow.rs:204-205` carries two strict inverses:

```rust
pub debug_topic_watchers: BTreeMap<OutputId, BTreeSet<Uuid>>,
pub debug_topic_subscriptions: BTreeMap<Uuid, BTreeSet<OutputId>>,
```

Two insert sites and two remove sites in `binaries/daemon/src/lib.rs:1820-1875`. Drift risk — adding one without updating the other silently desyncs.

## Fix options

1. Keep only `debug_topic_watchers` (hot-path lookup); iterate for unsubscribe — O(outputs) per unsubscribe is fine since unsubscribe is not hot
2. Wrap both maps in a small helper struct (`DebugTopicWatchers`) that exposes `insert/remove/watchers_for` and enforces the inversion invariant in one place

## Acceptance

- Single source of truth for the watcher-to-output relationship, with the other view derived or encapsulated

heyong4725:
## Summary

Closes #241. Drops `debug_topic_subscriptions: Uuid → Set<OutputId>` from the daemon's `RunningDataflow`. It was a strict inverse of `debug_topic_watchers: OutputId → Set<Uuid>`, maintained purely as a convenience for O(1) unsubscribe lookup.

Two insert sites and two remove sites is a drift risk — adding one without updating the other silently desyncs without compiler help. The #238 review flagged this, this PR closes the gap.

## Approach

Keep only `debug_topic_watchers` (it's the hot-path read on every node output dispatch in `send_topic_debug_frames` and on the `has_debug_watchers` check). For unsubscribe, scan with `retain`:

```rust
dataflow.debug_topic_watchers.retain(|_output_id, watchers| {
    watchers.remove(&subscription_id);
    !watchers.is_empty()
});
```

## Cost analysis

Unsubscribe fires once per CLI disconnect (not hot). Scan is bounded by the count of outputs with at least one active subscriber, which is bounded by the dataflow's total outputs (typically <100). In practice most outputs have zero subscribers, so the effective scan is short. Real cost is negligible; drift risk is gone.

## Test

`stop_debug_stream_scan_cleans_up_watchers` locks in the invariants the inverse map used to provide:

- Two subs (`sub_a`, `sub_b`) on `out_1`; only `sub_a` on `out_2`
- After stop for `sub_a`: `out_1` still has `sub_b` (entry retained), `out_2` is removed entirely (its only watcher was `sub_a`)

Mirrors the production retain pattern exactly. Any change that breaks the cleanup semantics fails this test.

## Test plan

- [x] `cargo test -p dora-daemon` — 28 passed, 0 failed
- [x] `cargo clippy --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python -- -D warnings` — clean

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1824,15 +1824,9 @@ impl Daemon {
             } => {
                 let result = if let Some(dataflow) = self.running.get_mut(&dataflow_id) {
                     for (node_id, data_id) in outputs {
-                        let output_id = OutputId(node_id.clone(), data_id.clone());
-                        dataflow
-                            .debug_topic_subscriptions
-                            .entry(subscription_id)
-                            .or_default()
-                            .insert(output_id.clone());
                         dataflow
                             .debug_topic_watchers
-                            .entry(output_id)
+                            .entry(OutputId(node_id, data_id))
                             .or_default()
                             .insert(subscription_id);
                     }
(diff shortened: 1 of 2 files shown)


heyong4725:
No findings.

I reviewed the daemon-side map collapse in binaries/daemon/src/lib.rs and binaries/daemon/src/running_dataflow.rs. The change is coherent: the hot-path OutputId -> Set<Uuid> map stays, unsubscribe now scans that single structure with retain, and I did not find any remaining runtime references that still depend on the removed inverse map.

Validation:
- cargo test -p dora-daemon

Residual risk: the new test locks in the unsubscribe cleanup semantics, but it still mirrors the production retain logic rather than driving the full StartTopicDebugStream / StopTopicDebugStream event path end-to-end.

PR #248, opened 2026-04-15
ci(nightly): smoke dora self update --check-only (#215 thread D read path)

heyong4725:
## Summary

Closes #215 thread D for the read path. Adds a nightly smoke step for `dora self update --check-only`. The destructive binary-swap path stays uncovered (needs a dedicated sandbox harness) and is documented as such in `testing-matrix.md`.

## What the step checks

1. **Entry banner prints** — `"Checking for updates"` must appear. If missing, the flag parsing or the code path entry is broken.
2. **One of two valid outcomes prints** — either `"latest version"` (already current) or `"update is available"` (current < upstream). Either is success.
3. **Rate-limit tolerance** — GHA runner IPs are shared and can hit GitHub's 60 req/hr unauthenticated limit. If neither outcome prints but rate-limit markers (`403`/`429`/"rate limit") appear, emit a warn and pass the step. A missing-outcome *without* rate-limit markers is treated as a real regression in the `self_update` crate's API contract.

## Local validation

```
$ dora self update --check-only
Checking for updates...
An update is available: 0.5.0. Run 'dora self update' to update
```

Matches the assertions.

## What this catches

- CLI flag parsing regressions (the `--check-only` flag)
- Binary builds with a broken `self_update` dependency
- GitHub API response format changes that break `get_latest_release()` (the `self_update` crate would panic or return an unexpected error)
- Silent "no output" regressions from transport errors

## What this does not catch

- Destructive `dora self update` (actual binary swap) — needs sandbox harness, deferred. Documented in `testing-matrix.md` under "NOT covered".

## Test plan

- [x] Local repro against live GitHub API
- [ ] Nightly run completes with new step green
- [ ] 3 consecutive green nightlies → eligible for Tier 0 promotion (follow-up)

PR #243, opened 2026-04-15
ci(nightly): topic echo/hz smoke + #236 regression guard

heyong4725:
## Summary

Part of #215 thread A (CI smoke for topic echo/hz/info on the WS inspection path, unblocked by #238).

- Add `dora topic echo --count 5` and `dora topic hz --duration 3` steps to the nightly `topic-and-top-smoke` job
- Strengthen the existing `dora topic info --duration 3` step to assert `Total messages >= 10` on a 10Hz fixture — the pre-#238 bug passed this step with `Total messages: 0` because zenoh silently dropped frames
- Update the job's header comment to reflect the WS inspection architecture
- Update `docs/testing-matrix.md`: remove echo/hz from "NOT covered", add them to `topic-and-top-smoke` coverage list with the regression-guard assertions

## Why nightly and not Tier 0

The WS inspection path is new (just landed in #238). Nightly lets us catch regressions without adding a new ~2-minute job to every push. Promote to Tier 0 after 3 consecutive green nightlies per the existing promotion policy (`docs/testing-matrix.md:176`).

## Validation (local)

Ran against `tests/fixtures/zenoh-debug-dataflow.yml` with the merged #238 CLI:

```
=== echo ===
{"timestamp":..,"name":"ticker/count","data":[29],"metadata":null}
{"timestamp":..,"name":"ticker/count","data":[30],"metadata":null}
{"timestamp":..,"name":"ticker/count","data":[31],"metadata":null}
=== hz ===
topic            avg_ms  avg_hz  min_ms  max_ms  std_ms  samples
ALL              100.04  10.00   77.32   124.61  6.56    30
ticker/count     100.04  10.00   77.32   124.61  6.56    30
=== info ===
Statistics:
  Total messages: 30
```

All three assertions pass.

## Test plan

- [x] Local repro against live coordinator+daemon with #238 CLI
- [ ] Nightly run completes with new steps green
- [ ] 3 consecutive green nightlies → eligible for Tier 0 promotion (follow-up)

PR #257, opened 2026-04-15
test(qa): cpu_affinity end-to-end regression test (#252)

### LINKED ISSUE #252: qa: cpu_affinity end-to-end test (affinity mask actually applied)
**Parent:** #250

## Gap

`cpu_affinity` in dataflow YAML is parsed (`libraries/message/src/descriptor.rs`) and passed to the spawn code (`binaries/daemon/src/lib.rs`), but no test verifies that the spawned node process actually lands with the requested CPU mask. A regression that drops the `sched_setaffinity` call or passes the wrong mask would go undetected.

## Proposed test

Add a Linux-only smoke test (integration or nightly):

1. Build a small Rust node that reads its own affinity via `sched_getaffinity(0, ...)` and emits the mask as output.
2. Fixture dataflow pins the node to a specific CPU subset (e.g. `cpu_affinity: [0, 1]`).
3. `dora run` the fixture with `--stop-after 2s`.
4. Assert captured output mask equals the requested mask.

## Scope

- Linux only (`sched_getaffinity` / `sched_setaffinity` are Linux-specific APIs)
- Single-daemon, single-node is enough — pinning is per-node config
- macOS/Windows: document as "feature not supported on non-Linux" in `docs/testing-matrix.md` (may already be the case — verify)

## Acceptance

- Nightly (or Tier 0 Linux-only) job runs the fixture and asserts the observed CPU mask matches the requested mask.
- Failure mode: if `cpu_affinity` ever stops applying, the test fails with a clear diff between expected and observed mask.

heyong4725:
## Summary

Closes #252. Adds a real regression test for `cpu_affinity` — proves the `sched_setaffinity` call at `binaries/daemon/src/spawn/prepared.rs:438` actually lands the requested CPU mask on the spawned process.

## What's added

- **`examples/cpu-affinity-probe/`** — small Rust binary that reads its own affinity mask via `sched_getaffinity(0, ...)` and prints `AFFINITY_MASK:<csv>` to stdout. On non-Linux it prints `AFFINITY_MASK:unsupported` (graceful skip, so the probe itself builds everywhere).
- **`examples/cpu-affinity-probe/dataflow.yml`** — fixture that pins the probe to cores `[0, 1]`.
- **`.github/workflows/nightly.yml`** — new `cpu-affinity-smoke` job (Linux-only, ~5 min). Runs the fixture, greps stdout for the marker line, asserts mask == `"0,1"`. Fails with a clear diff if the pin stops applying.
- **`docs/testing-matrix.md`** — entry in Tier 1 table.

## Validation (local, macOS)

```
$ dora run dataflow.yml --stop-after 3s
...
15:34:21 stdout  probe:  AFFINITY_MASK:unsupported
15:34:21 INFO    probe: daemon  probe finished successfully
```

Works end-to-end: probe spawns, reads affinity, prints marker, exits cleanly. On macOS the output is "unsupported" (by design). The nightly runs on ubuntu-latest where the probe returns the actual mask.

## Why nightly

Same rationale as #243, #247, #248: this is a Linux-only integration test with a real `cargo build` dependency, adding it to every PR push isn't worth ~1 min/run. Promotion to Tier 0 eligible after 3 consecutive green nightlies per the existing policy.

## What this catches

- Regressions in the `pre_exec` hook (`binaries/daemon/src/spawn/prepared.rs:427`) that compute the wrong mask
- Regressions that skip the `sched_setaffinity` call entirely
- Regressions in YAML parsing that drop the `cpu_affinity` field
- Platform-specific bugs if someone accidentally wraps the call in `cfg(target_os)` incorrectly

## What this does not catch

- macOS/Windows: neither OS has `sched_setaffinity`; the feature is Linux-only. The probe gracefully reports "unsupported" so the code runs without error, but actual pinning is not tested there.
- CPU reservation enforcement: this test only checks the process's *allowed* CPU set, not that the scheduler actually prefers those cores under load.

## Test plan

- [x] `cargo build -p cpu-affinity-probe` — clean
- [x] Local `dora run examples/cpu-affinity-probe/dataflow.yml --stop-after 3s` on macOS — probe runs, prints "unsupported" marker
- [ ] Nightly run on Linux completes with mask == `0,1`

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -485,10 +485,53 @@ jobs:
           dora stop --name tui-smoke --grace-duration 5s 2>/dev/null || true
           dora destroy 2>/dev/null || true
 
+  cpu-affinity-smoke:
+    # Regression test for #252. Spawns `cpu-affinity-probe` pinned to cores
+    # [0, 1] via the dataflow YAML; the probe reads its own affinity mask
+    # via sched_getaffinity and prints it. Test asserts the mask matches
+    # the requested set. Linux-only (sched_{get,set}affinity are Linux APIs).
+    name: cpu_affinity end-to-end (Linux)
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Run cpu-affinity-probe fixture (cpu_affinity: [0, 1])
+        run: |
+          cd examples/cpu-affinity-probe
+          out=$(dora run dataflow.yml --stop-after 3s 2>&1)
+          echo "$out"
+          # The probe prints exactly one `AFFINITY_MASK:<csv>` line from stdout.
+          # `dora run` prefixes node stdout with "stdout probe:" so extract the
+          # csv tail.
+          mask=$(echo "$out" | grep -oE 'AFFINITY_MASK:[^[:space:]]+' | head -1 | cut -d: -f2)
+          if [ -z "$mask" ]; then
+            echo "ERROR: probe did not print AFFINITY_MASK — cpu-affinity-probe failed to run"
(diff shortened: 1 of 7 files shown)


heyong4725:
1. .github/workflows/nightly.yml:586-593 updates file-issue-on-failure.needs to include cpu-affinity-smoke, but the created issue body’s Jobs: list still omits it. If this new job is the one that fails, the nightly regression issue will point maintainers at an incomplete job list and make the failure look unrelated. The new smoke job should be added there too so the reporter stays in sync with the actual nightly matrix.

I did not find a second concrete regression in the probe itself. Validation:
- cargo build -p cpu-affinity-probe

heyong4725:
Addressed in c965a2e7 — added `cpu-affinity-smoke` to the issue body's Jobs list alongside updating `needs:`. Maintainers hitting a failure will now see the right job listed.

Issue #255, opened 2026-04-15
qa: dataflow state reconstruction on coordinator restart


heyong4725:
**Parent:** #250

## Gap

README claims "dataflow state reconstruction on coordinator restart." This is the top-level integration of two lower-level features:

- redb-backed coordinator store (sibling: #250 Bucket A item)
- daemon auto-reconnect loop (sibling: #250 Bucket A item)

The combined end-to-end claim — that a running dataflow survives a coordinator restart with no user-visible disruption — has no test.

## Proposed test

Nightly smoke:

1. Start coordinator with redb backend, daemon, fixture dataflow with steady output
2. Record current state: dataflow UUID, node list, recent output timestamps
3. `pkill` coordinator — do not touch daemon
4. Restart coordinator with same redb store path
5. Wait for daemon to reattach (within bounded window)
6. Assert: same dataflow UUID present, same node list, output continues from roughly where it left off (not from 0)

## Dependencies

This test will fail cleanly if either sibling is broken, which is useful diagnostic behavior. Land the two prerequisite tests first so failures here have attribution.

- Sibling: redb-backend E2E
- Sibling: daemon auto-reconnect E2E

## Acceptance

- Nightly job passes.
- State reconstruction claim in README is backed by an automated test rather than documentation.

Issue #254, opened 2026-04-15
qa: daemon auto-reconnect end-to-end test


heyong4725:
**Parent:** #250

## Gap

Partial coverage exists: #242 landed `restore_topic_debug_streams_re_issues_start_after_reconnect`, which tests the topic-debug-stream re-registration path when a daemon reconnects. The **daemon-level reconnect loop itself** — heartbeat timeout detection → retry with exponential backoff → successful reattach — is not end-to-end tested.

## Proposed test

Integration test under `tests/` (not nightly-only — reconnect is a load-bearing claim):

1. Start coordinator + daemon + small fixture dataflow with continuous output (Python ticker fixture works)
2. Verify topic echo sees frames
3. `pkill -STOP` the daemon process (freeze, don't kill — simulates network partition)
4. Wait past heartbeat timeout
5. `pkill -CONT` the daemon (resume)
6. Assert daemon shows up in `dora list` as active again
7. Assert topic echo starts receiving frames again after some bounded window

## Constraints

- Linux-only (SIGSTOP/SIGCONT) — document in `testing-matrix.md` as Linux scope
- Tricky: getting heartbeat timeout + backoff tuning right to make this reliable under GHA runner load. May need to tune heartbeat interval via a test-only env var or config.

## Acceptance

- Reconnect test lives in `tests/fault-tolerance-e2e.rs` or a new `tests/daemon-reconnect-e2e.rs`
- Flakiness budget: <1% false-positive rate over a 2-week nightly window before promotion to Tier 0

PR #258, opened 2026-04-15
test(qa): redb backend end-to-end smoke (#253)

### LINKED ISSUE #253: qa: redb-backend end-to-end smoke (state survives coordinator restart)
**Parent:** #250

## Gap

`libraries/coordinator-store/src/redb_store.rs` has strong unit coverage (10+ tests: CRUD, schema versioning, persistence). What's missing: an end-to-end test that proves the redb backend actually survives a coordinator restart with real dataflows.

## Proposed test

Nightly smoke job:

1. Start `dora coordinator --backend redb --store-path /tmp/dora-redb-smoke`
2. Start daemon, start a small fixture dataflow (`--detach`), record its UUID
3. `pkill -f "dora coordinator"` — simulate coord crash
4. Restart `dora coordinator --backend redb --store-path /tmp/dora-redb-smoke`
5. Assert `dora list` shows the same dataflow UUID with the same status after restart
6. Teardown

## Why it matters

The README claims "persistent redb-backed state store, daemon auto-reconnect, dataflow state reconstruction on coordinator restart" as a first-class feature. Unit tests prove the store layer is correct in isolation; this test proves the integration actually delivers on the README claim.

## Related

- Depends on daemon auto-reconnect working (see sibling issue). If the coordinator comes back but daemons never reattach, the test fails — and the failure correctly identifies "reconnect is broken."
- State reconstruction (sibling issue) is the next layer up from this one.

## Acceptance

- Nightly job passes consistently.
- A regression where redb store is corrupted/not read on restart fails the test with a clear `dora list` diff.

heyong4725:
## Summary

Closes #253. Adds a nightly end-to-end smoke test that proves the redb coordinator store survives a coord restart — the capability the README advertises as "persistent redb-backed state store."

## Design: what the test actually proves

Two sessions wrapping a coord restart:

### Session 1 — writes
- Start coord with `--store redb:/tmp/dora-redb-smoke.db` + daemon
- Run a short dataflow (cpu-affinity-probe fixture, finishes in ~1s)
- **Assertion:** `strings $STORE | grep -q "redb-smoke"` — the dataflow name appears in the redb file as evidence of actual writes (not just preallocation — redb creates a 1.5MB file on init regardless of usage)
- SIGTERM the coord, leave daemon orphaned

### Session 2 — reads
- Restart coord against the same store path
- **Assertion:** session 2 coord log contains `cleared N stale daemon records` — that code path only fires if the restart successfully read at least one daemon record from redb. End-to-end round-trip proof.
- **Assertion:** no `corrupt`, `panicked`, or `ERROR` markers in session 2 log

## Why this signal

During local testing, the `"cleared 1 stale daemon records"` line from `binaries/coordinator/src/lib.rs` turned out to be the reliable observable end-to-end indicator:

- Store size doesn't change meaningfully across a short dataflow (preallocation absorbs the delta)
- `dora list` after restart doesn't show Succeeded/Failed dataflows (known limitation: succeeded records aren't re-hydrated into `running_dataflows`, only left in the store)
- Log-level signals at restart are the cleanest semantic proof

## What's not tested

- Full state reconstruction for Running dataflows (covered by sibling #255)
- Daemon auto-reconnect loop (covered by sibling #254)
- The Finished-dataflow hydration into `dora list` — that's a feature gap, not a test bug. `ControlRequest::List` at `lib.rs:973` only reads in-memory state; the store records for Finished dataflows are inaccessible via CLI today. Worth its own issue if surfaced.

## Test plan

- [x] Local validation: coord1 writes dataflow record, coord2 logs `"cleared 1 stale daemon records"` and starts cleanly
- [x] Pre-existing redb unit tests still pass (`cargo test -p dora-coordinator-store`)
- [ ] Nightly runs pass on GHA Linux runner

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -528,10 +528,87 @@ jobs:
           fi
           echo "OK: probe ran with the expected affinity mask: $mask"
 
+  redb-backend-smoke:
+    # Regression test for #253. Proves the redb coordinator store survives
+    # a coord restart: writes during normal operation, reads cleanly on
+    # startup. The load signal is a tracing line the restarted coord prints
+    # after reading stored daemon records — see binaries/coordinator/src/lib.rs
+    # `"cleared N stale daemon records"`.
+    name: redb backend end-to-end
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Session 1 — write state to redb
+        run: |
+          STORE=/tmp/dora-redb-smoke.db
+          rm -f "$STORE"
+          # Start coord with persistent redb store + daemon.
+          dora coordinator --store "redb:$STORE" > /tmp/coord1.log 2>&1 &
+          sleep 2
+          dora daemon > /tmp/daemon1.log 2>&1 &
+          sleep 2
+          # Run a short dataflow so the coord writes dataflow + daemon records
+          # to the store.
+          dora start examples/cpu-affinity-probe/dataflow.yml --name redb-smoke --detach
+          sleep 6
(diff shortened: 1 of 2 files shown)


heyong4725:
No findings.

I reviewed the .github/workflows/nightly.yml smoke job and the docs/testing-matrix.md update. The new job is internally consistent:
- it writes real state into a redb-backed coordinator
- it verifies a dataflow write signal in the store file
- it then restarts the coordinator against the same store and checks for the stale-daemon cleanup log that only appears after persisted daemon records are read back

I also re-checked the coordinator startup path for that log signal in binaries/coordinator/src/lib.rs, and ran:
- cargo test -p dora-coordinator-store

Residual risk: this still proves persistence via observable write/read signals rather than by exercising a user-facing recovered-state CLI path, so it won’t catch gaps in future “rehydrate finished dataflows into dora list” behavior. For the PR’s stated scope, that looks reasonable.

PR #259, opened 2026-04-15
test(qa): daemon auto-reconnect end-to-end smoke (#254)

### LINKED ISSUE #254: qa: daemon auto-reconnect end-to-end test

heyong4725:
## Summary

Closes #254. End-to-end regression test for the daemon auto-reconnect loop — the piece that catches "daemon silently drops off and never comes back" bugs that unit tests can't see.

## Flow

1. Start coord + daemon, assert initial registration (coord logs `daemon X reports N running dataflow(s)`)
2. `kill -STOP $DAEMON_PID` — simulates network partition or hung process
3. Sleep 35s — coord's watchdog timeout is 30s at `binaries/coordinator/src/lib.rs:1664`
4. Assert coord logged `Disconnecting daemons that failed watchdog`
5. `kill -CONT $DAEMON_PID` — daemon's WS send fails immediately, it enters the reconnect loop (first retry: 1s backoff)
6. Sleep 30s — generous margin for scheduler jitter
7. Assert daemon logged `daemon disconnected from coordinator` (proves it detected the drop)
8. Assert coord saw >= 2 daemon status reports (initial + re-registered)

Total runtime ~70s per nightly run.

## Local validation

Ran the exact same flow on macOS (SIGSTOP/SIGCONT are POSIX, nightly runs on Linux):

```
OK: registered
frozen
OK: watchdog fired
resumed
OK: daemon detected disconnect
coord saw 2 reports
```

All 4 assertions pass.

## Why Linux-only

The test design relies on process-freeze semantics that are POSIX but the implementation paths differ enough across platforms that I'm not gating it cross-platform. macOS would work; Windows would not (no SIGSTOP). Nightly runs on ubuntu-latest, which matches.

## What this catches

- Regressions in the daemon's reconnect loop at `binaries/daemon/src/lib.rs:470-552` (wrong backoff, silent retry bailout, missing re-register)
- Regressions in the coord's watchdog at `binaries/coordinator/src/lib.rs:1655-1710` (timeout threshold change, disconnect cleanup gap)
- Bugs that show up only when the daemon reappears after the coord has already removed it from `daemon_connections`

## What this does not catch

- Subtle state reconciliation bugs after reconnect (e.g., does the daemon correctly resubscribe to active topic streams?). That's #242's `restore_topic_debug_streams_re_issues_start_after_reconnect` territory for topic streams; for other subsystems, it's future work.
- Multi-daemon reconnect scenarios.

## Test plan

- [x] Local repro on macOS — all 4 assertions pass
- [x] `cargo fmt --all -- --check` — n/a (YAML only)
- [x] `cargo clippy --all ... -D warnings` — n/a (YAML only)
- [ ] Nightly runs pass on GHA Linux runner

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -605,10 +605,96 @@ jobs:
         run: |
           pkill -f "dora (daemon|coordinator)" 2>/dev/null || true
 
+  daemon-reconnect-smoke:
+    # Regression test for #254. Proves the daemon auto-reconnect loop works
+    # end-to-end by SIGSTOPping the daemon long enough for the coord's
+    # heartbeat watchdog to disconnect it (30s timeout in coord/src/lib.rs:1664),
+    # then SIGCONTing and asserting the daemon rediscovers + re-registers.
+    # Linux-only: relies on SIGSTOP/SIGCONT semantics for freezing the process.
+    name: daemon auto-reconnect end-to-end (Linux)
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Start coord + daemon, capture PID
+        run: |
+          dora coordinator > /tmp/coord.log 2>&1 &
+          echo "COORD_PID=$!" >> "$GITHUB_ENV"
+          sleep 2
+          dora daemon > /tmp/daemon.log 2>&1 &
+          echo "DAEMON_PID=$!" >> "$GITHUB_ENV"
+          sleep 3
+          # Sanity: daemon registered and emitted its first status report.
+          if ! grep -q "daemon.*reports.*running dataflow" /tmp/coord.log; then
+            echo "ERROR: daemon did not register — no 'daemon X reports' line in coord log"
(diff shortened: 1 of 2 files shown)


heyong4725:
No findings.

I reviewed the .github/workflows/nightly.yml smoke job and the docs/testing-matrix.md update against the current coordinator watchdog and daemon reconnect paths. The test logic is internally consistent with the implementation:
- coordinator watchdog disconnects after > 30s
- daemon detects lost coordinator heartbeat after > 20s once resumed
- the chosen log markers match the actual code paths for disconnect and re-registration

I did not find a concrete workflow bookkeeping mismatch either; the nightly failure-reporting job was updated consistently with the new daemon-reconnect-smoke job.

Residual risk: this is still a timing-sensitive integration test driven by fixed sleeps, so the remaining risk is CI flakiness under heavy runner load rather than a correctness issue in the assertions themselves.

Issue #260, opened 2026-04-15
daemon: kills running dataflows on coord WS drop — breaks state reconstruction claim


heyong4725:
**Parent:** #250 (follow-up surfaced while implementing #255)

## Symptom

README and `guide/src/introduction.md` describe "dataflow state reconstruction on coordinator restart" as a feature. In practice, when the coordinator disconnects, running dataflows are **killed on the daemon side**, not reclaimed across restart.

## Root cause

`binaries/daemon/src/running_dataflow.rs:157-163` — the `ProcessHandle::drop` impl submits `ProcessOperation::Kill` and logs `"process was killed on drop because it was still running"`. When the daemon's WS connection to the coordinator drops, the daemon's run loop exits and the owning structs are dropped, cascading through `ProcessHandle::drop` → node processes SIGKILLed.

Observed in a local repro while implementing #255:

```
daemon log:
  WARN running_dataflow: process was killed on drop because it was still running
  ERROR spawn::prepared: node exited with error: Signal(9)
       node_id=source dataflow_id=019d9308-...
  WARN daemon: daemon disconnected from coordinator: ... Attempting reconnect...
  INFO daemon: reconnecting in 1s...

coord (session 2, after restart):
  INFO coordinator restarted: dataflow 019d9308-... -> Recovering
  INFO daemon 019d9308-6f9c-... reports 0 running dataflow(s)   ← 0, not 1
```

The daemon reconnects cleanly, but the dataflow it was running is gone. The coord's reconciliation path at `binaries/coordinator/src/lib.rs:1983` expects the daemon to report still-running dataflows; since the daemon killed them, no reconciliation happens. The Recovering record eventually transitions to Failed via the 60s timeout at `lib.rs:1731`.

## Why it matters

README HA claims set expectations that the implementation doesn't meet. A user who kicks off a long-running dataflow, then restarts the coordinator for any reason (config change, upgrade, crash), loses all in-flight work. No recovery.

## Scope

Two changes needed, probably in sequence:

1. **Daemon should keep running dataflows alive across coord WS drop.** The drop-cascade needs to be gated — don't kill nodes when the connection loss is "transient" (i.e., the reconnect loop is entered). Only kill if the reconnect truly fails terminally, or if the new coord doesn't claim the dataflow.
2. **Daemon should re-report running dataflows on reconnect.** The existing `DaemonStatusReport` flow already does this when the daemon has them — so once (1) is fixed, (2) may be free.

[...]

PR #261, opened 2026-04-15
test(qa): partial state reconstruction smoke + doc accuracy (#255)

### LINKED ISSUE #255: qa: dataflow state reconstruction on coordinator restart

heyong4725:
## Summary

Closes #255 for the scope actually achievable today. Discovered during implementation that the feature the README advertises ("dataflow state reconstruction on coordinator restart") is only partially implemented. Delivers the honest partial test + files a new issue for the real gap + fixes doc accuracy.

### 1. Partial nightly test

New `state-reconstruction-smoke` nightly job. Validates what the current implementation does deliver:

- **Session 1**: start coord (redb) + daemon + long-running dataflow. Assert the dataflow name appears in the redb store file via `strings $STORE | grep sr-test` — proves writes happen for Running dataflows.
- **Session 2**: kill coord, restart against the same store. Assert `coordinator restarted: dataflow X -> Recovering` log line — proves the startup hydration path at `binaries/coordinator/src/lib.rs:302` reads persisted Running records and transitions them to Recovering.

Local validation: both assertions pass in ~15s.

### 2. Design gap issue (#260)

During implementation I hit a surprise: when the coord WS disconnects, the daemon's `ProcessHandle::drop` at `binaries/daemon/src/running_dataflow.rs:157-163` kills the running nodes with SIGKILL. Evidence:

```
daemon log:
  WARN running_dataflow: process was killed on drop because it was still running
  ERROR spawn::prepared: node exited with error: Signal(9)

coord session 2:
  INFO coordinator restarted: dataflow ... -> Recovering
  INFO daemon X reports 0 running dataflow(s)   ← 0, not 1
```

The reconciliation path at `lib.rs:1983` would promote Recovering → Running if the daemon reported a still-alive dataflow. It doesn't. Eventually the Recovering record times out to Failed via the 60s watchdog.

Filed dora-rs/adora#260 with a repro and proposed scope. Not in this PR — that's daemon lifecycle work, separate surface area.

### 3. Docs updated to match reality

`README.md:49` and `guide/src/introduction.md:26` both said "dataflow state reconstruction on coordinator restart." Updated to:

> "dataflow records survive coordinator restart (full running-dataflow reclaim across restart is partial; see open tracker)"

Honest about what's implemented. When #260 lands, this reverts to the strong claim.

## Why not lock in the broken behavior via a test

Option C ("test what current-broken does, fail when someone fixes it") was considered. Rejected because it ties the test to a specific bug manifestation rather than the desired behavior. The partial test here stays green when the gap is fixed; the new assertion gets added then (see the TODO comment inline in the job).

## Test plan

- [x] Local repro on macOS — both assertions pass
- [x] Gap reproduced + documented in #260
- [x] `cargo fmt --all -- --check` — n/a (YAML + markdown only)
- [ ] Nightly run on GHA Linux picks up the new job

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -691,10 +691,95 @@ jobs:
           pkill -CONT -f "dora daemon" 2>/dev/null || true
           pkill -f "dora (daemon|coordinator)" 2>/dev/null || true
 
+  state-reconstruction-smoke:
+    # Partial regression test for #255. Validates what the current redb
+    # backend + recovery path actually delivers on coord restart: store
+    # records survive, and the coord transitions previously-Running
+    # dataflows to Recovering on startup.
+    #
+    # NOT tested here: daemons keeping running dataflows alive across
+    # coord disconnect. That part of the README claim ('dataflow state
+    # reconstruction on coordinator restart') is aspirational — see the
+    # follow-up issue for the implementation gap. When the daemon side
+    # lands, this job should be extended to assert the Recovering -> Running
+    # reconciliation path at binaries/coordinator/src/lib.rs:1996.
+    name: state reconstruction end-to-end (partial)
+    runs-on: ubuntu-latest
+    needs: build-cli
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Download dora CLI
+        uses: actions/download-artifact@v4
+        with:
+          name: dora-cli
+          path: ./cli-bin
+      - name: Install CLI to PATH
+        run: |
+          chmod +x ./cli-bin/dora
+          echo "$PWD/cli-bin" >> "$GITHUB_PATH"
+      - name: Pre-build example nodes
+        run: |
+          cargo build -p rust-dataflow-example-node
+      - name: Session 1 — run long-lived dataflow, kill coord
+        run: |
+          STORE=/tmp/state-reconstruct.db
+          rm -f "$STORE"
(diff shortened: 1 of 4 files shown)


heyong4725:
No findings.

I reviewed the new state-reconstruction-smoke job and the README/guide/testing-matrix wording updates. The job is honest about scope and the asserted log signal matches the current coordinator startup path: persisted Running records are read from redb and transitioned to Recovering, which is exactly what the workflow checks.

The doc changes also improve accuracy rather than over-claiming behavior. They now match the current implementation gap instead of promising full running-dataflow reclaim across coordinator restart.

Residual risk: the job still proves reconstruction via store/log signals rather than a user-facing recovered-state command path, so it won’t catch future regressions in any eventual CLI-visible “reclaimed running dataflow” surface. For the PR’s stated scope, that looks appropriate.

Issue #266, opened 2026-04-15
feat(descriptor): global env field on dataflow YAML (upstream #1341)


heyong4725:
Sub-issue of #201. Scope narrowed after review (see comment): only #1341 brings real value; #1359 and #1430 dropped as not-worth-the-churn.

## Scope

- [ ] **#1341** — Global `env:` field on dataflow YAML
  - `libraries/message/src/descriptor.rs` currently only supports per-node `env`.
  - Add a top-level `env` on `Descriptor` that is inherited by all nodes; per-node `env` overrides on key conflict.
  - Merge at descriptor resolution (`resolve_aliases_and_set_defaults`) so downstream spawn paths are unchanged.
  - Document precedence (global < per-node).

## Dropped (see comment)

- **#1359** collect-all validation — type checking already collects-all; structural errors cascade, collect-all multiplies noise. Refactor risk > UX win.
- **#1430** URL source validation — fetch-time errors are already clear; not worth adding a `url` workspace dep.

## Verification

- `cargo test -p dora-core -p dora-message`
- YAML fixture: global env inherits, per-node key wins, missing global env leaves per-node semantics unchanged.

## References

Upstream PR: dora-rs/dora#1341

heyong4725:
## Scope narrowed to #1341 (2026-04-15)

Re-examined each candidate against actual adora value.

### Keeping: #1341 — Global `env:` field

Real user pain. Shared env vars (`RUST_LOG`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `CUDA_VISIBLE_DEVICES`) currently require copy-paste into every node block; changing one value edits N places. Small implementation (add `Descriptor.env`, merge at resolve-time with per-node winning). No new deps. Matches upstream YAML schema — zero migration cost for user YAMLs.

### Dropping: #1359 — Collect-all validation

- Type checking **already** collects-all (`check_type_annotations_full` at `validate.rs:734`). The remaining scope is structural/path errors.
- Most YAML errors cascade (missing node → 3+ downstream symptoms). Collect-all often multiplies noise rather than reducing roundtrips.
- Cost: 41 `bail!` sites to convert to accumulator pattern, with regression risk in every validation path.
- Refactor risk > UX win.

### Dropping: #1430 — URL source validation

- Four TODO sites at `validate.rs:79, 106, 120, 127`. Today a malformed URL falls through to a fetch-time error with a clear message.
[...]

Issue #263, opened 2026-04-15
fix(arrow): checked_add for Arrow IPC buffer-offset bounds (upstream #1479)


heyong4725:
Sub-issue of #201. After re-examining each candidate (see comment below), only one is a real panic-safety bug in adora. The other three are dropped from scope.

## Scope

- [ ] **#1479** — Checked arithmetic for Arrow IPC buffer-offset bounds
  - `apis/rust/node/src/node/arrow_utils.rs:152` does `if offset + len > raw_buffer.len()` on `usize` fields received over the wire via `ArrowTypeInfo::buffer_offsets`.
  - A peer sending `offset = usize::MAX, len = 1` overflows the add to `0`, bypasses the check, then `raw_buffer.slice_with_length(usize::MAX, 1)` panics inside arrow.
  - Replace with `offset.checked_add(len)` and bail cleanly on overflow or out-of-bounds.
  - Add regression test with an `ArrowTypeInfo` containing `offset = usize::MAX` to ensure we return an error instead of panicking.

## Dropped from scope (see comment)

- **#1438 / #1440** — `to_str().unwrap()` on `CARGO_MANIFEST_DIR` is unreachable (path is baked in at compile time)
- **#1562** — All non-test `Uuid::parse_str` in adora already use `.ok()` or operate on internally-generated UUIDs; no log-field parsing path
- **#1430** — URL source validation is UX, not panic-safety; belongs with #266

## Verification

- `cargo test -p dora-node-api`
- `cargo clippy --all -- -D warnings`

## References

Upstream PR: dora-rs/dora#1479

heyong4725:
## Scope reassessment (2026-04-15)

Re-examined each item critically before starting implementation. Three of the four don't apply to adora:

### Dropping from scope

- **#1438 / #1440** — Non-UTF-8 path in `to_str().unwrap()` on `CARGO_MANIFEST_DIR` only fires when `use_path_deps=true`. `CARGO_MANIFEST_DIR` is baked in at compile time — if the build path weren't UTF-8, the binary wouldn't exist. Effectively unreachable. Not worth the churn.

- **#1562** — Audited every non-test `Uuid::parse_str` in adora:
  - `binaries/cli/{common.rs,command/node/list.rs,command/trace/view.rs}` — all use `.ok()` / `.is_ok()`, safe
  - `libraries/message/src/common.rs` — all `.ok()`, safe
[...]

Issue #201, opened 2026-04-13
Incorporate upstream changes


phil-opp:
According to Claude, the following things changed in the upstream dora-rs/dora repo since creating the fork:

Tracking issue for features merged into [dora-rs/dora](https://github.com/dora-rs/dora) that have **not** been ported to adora. Ordered roughly by impact.

### Architecture / Transport

- [ ] **tarpc RPC for coordinator-daemon communication**
  [#1372](https://github.com/dora-rs/dora/pull/1372), [#1345](https://github.com/dora-rs/dora/pull/1345), [#1367](https://github.com/dora-rs/dora/pull/1367)
  Replaces WebSocket-based coordinator↔daemon protocol with tarpc. Includes a `get_version` RPC for CLI-coordinator version checking.
  - Note: we might want to keep websocket

- [ ] **Zenoh-based log relay (replace TCP log relay)**
  [#1380](https://github.com/dora-rs/dora/pull/1380), [#1554](https://github.com/dora-rs/dora/pull/1554)
  Replaces TCP-based log forwarding with zenoh publish/subscribe. Adds per-node subtopics, historical log buffering, and deduplication. Adora still has no zenoh log publisher in the daemon.

- [ ] **Replace flume with tokio mpsc in EventStream (deadlock fix)**
  [#1603](https://github.com/dora-rs/dora/pull/1603)
  Fixes a deadlock by switching `EventStream` internals from `flume` to `tokio::sync::mpsc`. Adora still uses `flume` throughout `apis/rust/node/src/event_stream/`.

- [x] **Remove unix domain socket communication option**
  [#1599](https://github.com/dora-rs/dora/pull/1599)
  Cleanup of unused UDS transport path. Need to verify if adora still carries this code.

- [ ] **Machine UID detection for same-machine checks**
  [#1382](https://github.com/dora-rs/dora/pull/1382)
  Uses `machine-uid` crate instead of IP comparison for reliable same-machine detection. Not present in adora.

### Dataflow / Node Features

- [ ] **Hot-reload for custom node binaries**
  [#1328](https://github.com/dora-rs/dora/pull/1328)
  Adds file-watching and live-reload for user-built node binaries (not just Python operators). Adora has operator reload but not custom binary hot-reload.

- [ ] **Global `env` field on YAML dataflow files**
  [#1341](https://github.com/dora-rs/dora/pull/1341)
[...]

PR #268, opened 2026-04-15
fix(arrow): checked_add for Arrow IPC buffer-offset bounds (upstream #1479)

heyong4725:
## Summary

- Replace unchecked `offset + len` with `checked_add` in Arrow IPC buffer-offset bounds check (`apis/rust/node/src/node/arrow_utils.rs`)
- Add regression test exercising `offset = usize::MAX`

## Why

`buffer_into_arrow_array_inner` validates wire-received `BufferOffset` values against the raw buffer length:

```rust
if offset + len > raw_buffer.len() {
```

`offset` and `len` are `usize` fields in `ArrowTypeInfo::buffer_offsets`, received from another node. A peer sending `offset = usize::MAX, len = 1` overflows the add to `0`, passes the check, then `raw_buffer.slice_with_length(usize::MAX, 1)` panics inside arrow — aborting the node process.

The fix uses `checked_add` and returns an eyre error on overflow, same pattern the file already uses for the other wire-bounds checks (`MAX_TYPE_DEPTH`, `MAX_ENTRIES`, `MAX_IPC_BYTES`).

Backport of upstream [dora-rs/dora#1479](https://github.com/dora-rs/dora/pull/1479). Refs #263, #201.

### Scope note

#263 originally batched four upstream PRs. After auditing, three don't apply to adora and were dropped with rationale (see #263 comment):
- #1438 / #1440 — `to_str().unwrap()` on `CARGO_MANIFEST_DIR` is unreachable (compile-time path)
- #1562 — all non-test `Uuid::parse_str` in adora already `.ok()`-based; no log-field parse path
- #1430 — URL source validation is UX, not panic-safety; folded into #266

## Test plan

- [x] `cargo test -p dora-node-api` — 7/7 arrow_utils tests pass, including new `buffer_into_arrow_array_rejects_overflow_offset`
- [x] `cargo clippy -p dora-node-api -- -D warnings`
- [ ] CI green on Ubuntu/macOS/Windows

### DIFF
--- a/apis/rust/node/src/node/arrow_utils.rs
+++ b/apis/rust/node/src/node/arrow_utils.rs
@@ -149,7 +149,13 @@ fn buffer_into_arrow_array_inner(
 
     let mut buffers = Vec::new();
     for BufferOffset { offset, len } in &type_info.buffer_offsets {
-        if offset + len > raw_buffer.len() {
+        // Use checked_add to guard against malicious/buggy peers sending
+        // `offset` or `len` values that would overflow `usize` on addition
+        // and bypass the bounds check below.
+        let end = offset
+            .checked_add(*len)
+            .ok_or_else(|| eyre::eyre!("buffer offset overflow: offset={offset}, len={len}"))?;
+        if end > raw_buffer.len() {
             eyre::bail!(
                 "buffer offset out of bounds: offset={offset}, len={len}, buffer_len={}",
                 raw_buffer.len()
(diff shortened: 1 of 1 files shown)


heyong4725:
No findings.

I reviewed the single-file fix in apis/rust/node/src/node/arrow_utils.rs. Replacing `offset + len` with `checked_add` is the correct hardening for peer-controlled `BufferOffset` values, and the new regression test covers the exact overflow case that previously could bypass the bounds check and panic later inside Arrow.

Validation:
- cargo test -p dora-node-api

Residual risk: this closes the specific integer-overflow panic on buffer slicing, but it doesn’t broaden the wire-format validation model beyond the existing per-field guards. For this PR’s scope, that looks correct.

Issue #265, opened 2026-04-15
fix(event-stream): replace flume with tokio mpsc to fix deadlock (upstream #1603)


heyong4725:
Sub-issue of #201. Backport upstream deadlock fix.

## Problem

`apis/rust/node/src/event_stream/` uses `flume` channels. Upstream dora-rs/dora#1603 identified a deadlock in this path and migrated to `tokio::sync::mpsc`. Adora is still on `flume`.

## Scope

- [ ] Replace `flume::Receiver` / `flume::Sender` with `tokio::sync::mpsc` in the EventStream internals
  - `apis/rust/node/src/event_stream/mod.rs:68` (`flume::r#async::RecvStream`)
  - `apis/rust/node/Cargo.toml:24` (drop flume dep if no longer needed)
- [ ] Update `Stream` impl to wrap `tokio_stream::wrappers::ReceiverStream` (or equivalent)
- [ ] Verify drop-stream / merged-stream paths still compose cleanly

## Verification

- Reproduce the upstream deadlock scenario if possible, confirm fix
- `cargo test -p dora-node-api`
- Smoke tests: `./scripts/smoke-all.sh --rust-only`
- Full E2E: `cargo test --test ws-cli-e2e` and `cargo test --test fault-tolerance-e2e`

## References

Upstream PR: dora-rs/dora#1603

heyong4725:
## Scope confirmation + narrowing (2026-04-15)

Verified against adora that the upstream deadlock (dora-rs/dora#1603) is **real and latent**, not speculative:

- `dora-node-api` uses `flume 0.10.14` directly (`cargo tree -i flume:0.10.14`); the vulnerable `flume::r#async::RecvStream` is at `apis/rust/node/src/event_stream/mod.rs:68`.
- `pyo3 0.28.3/src/coroutine/waker.rs:43-51` still acquires the Python GIL inside `wake_by_ref` via `Python::attach`. The pyo3 upgrade did not change this behavior.
- Adora's Python async path exists: `recv_async` at `apis/python/node/src/lib.rs:286`, examples `python-async` and `python-recv-async`, test `test_mock_node_recv_async`.

All three preconditions hold → same AB-BA deadlock applies.

### Narrowed scope

`flume` is used in 55 files across the workspace (daemon, runtime, ROS2 bridge, other node paths). None of those involve the Python GIL; they don't deadlock. Scope creeps into unrelated churn if we touch them.

[...]

PR #269, opened 2026-04-15
fix(event-stream): replace flume with tokio mpsc to fix GIL deadlock (upstream #1603)

heyong4725:
## Summary

Swap the EventStream's `EventItem` channel from `flume` to `tokio::sync::mpsc` to close the AB-BA deadlock described in upstream [dora-rs/dora#1603](https://github.com/dora-rs/dora/pull/1603).

## Why

**The deadlock recipe:**
- `flume 0.10.14`'s `AsyncSignal::fire()` holds `Spinlock<Waker>` while calling `waker.wake_by_ref()`.
- `pyo3 0.28.3`'s `AsyncioWaker::wake_by_ref` (at `pyo3-0.28.3/src/coroutine/waker.rs:43-51`) acquires the Python GIL inside `Python::attach`.
- Event-stream thread sends → holds Spinlock, waits for GIL.
- Python task polls `RecvStream` → holds GIL, waits for Spinlock.
- AB-BA deadlock.

**All three preconditions hold in adora:**
- `flume 0.10.14` is a direct dep of `dora-node-api` (`cargo tree -i flume:0.10.14`).
- `pyo3 0.28` still GILs inside the waker (I read the source — the upgrade did not fix this).
- Python async path is live: `recv_async` is exposed at `apis/python/node/src/lib.rs:286`; examples `python-async` and `python-recv-async` exercise it.

## Scope

Narrow on purpose. `flume` is used in 55 files across the workspace; only the EventStream's `EventItem` channel feeds the GIL-acquiring waker. Other flume channels (ack, drop signaling, thread-join) stay — they run on blocking threads with no async waker involvement.

Touched:
- `apis/rust/node/src/event_stream/thread.rs` — producer: `mpsc::Sender<EventItem>`, `blocking_send`, updated error match.
- `apis/rust/node/src/event_stream/mod.rs` — consumer: receiver field is raw `tokio::sync::mpsc::Receiver<EventItem>` (no stream wrapper — keeps `is_empty` direct); call sites: `.recv()` / `.try_recv()` / `.poll_recv(cx)` / `.is_empty()`. Zenoh subscriber threads also use `blocking_send`.
- `apis/rust/node/Cargo.toml` — add `sync` feature to tokio dep.
- `Cargo.toml` — bump workspace tokio minimum `1.24.2` → `1.28` (needed for `mpsc::Receiver::is_empty`, exposed through `EventStream::is_empty`). Lockfile already resolves 1.51.x; this just tightens the declared MSV.

Not touched: `ack_channel: flume::Sender<()>`, `drop_channel`, `EventStreamThreadHandle` internal join channel, drop-token pending receivers, daemon/runtime/bridge's own flume usage.

## Test plan

- [x] `cargo build -p dora-node-api` — clean
- [x] `cargo test -p dora-node-api` — 76 unit + 3 event_stream + 4 doctests pass
- [x] `cargo test --all --exclude <python>` — all workspace tests pass
- [x] `cargo clippy --all --exclude <python> -- -D warnings` — clean
- [ ] Manually run `python-recv-async` example under load to confirm no hang (optional; the fix is structural and test coverage is functional)
- [ ] CI green on all three platforms

## Notes on verification

No regression test was added for the deadlock itself — reproducing the AB-BA lock requires a real Python coroutine polling the receiver concurrently with a rust-thread sender, both going through pyo3's GIL-acquiring waker. That's a full integration test against the live binding, not a unit test. The functional unit tests (76 passing) cover the channel's functional behavior; the deadlock fix is structural and verified by code review.

Refs #265, #201.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ serde_json = "1.0.145"
 thiserror = "2.0"
 colored = "3.0.0"
 inquire = { version = "0.7.5", default-features = false }
-tokio = { version = "1.24.2", default-features = false }
+tokio = { version = "1.28", default-features = false }
 eyre = "0.6.8"
 serde = { version = "1", features = ["derive"] }
 tracing = "0.1.36"
--- a/apis/rust/node/Cargo.toml
+++ b/apis/rust/node/Cargo.toml
@@ -35,7 +35,7 @@ futures-timer = "3.0.2"
 dora-arrow-convert = { workspace = true }
 aligned-vec = "0.5.0"
 serde_json = { workspace = true, features = ["preserve_order"] }
-tokio = { workspace = true, features = ["rt-multi-thread", "time", "macros"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "time", "macros", "sync"] }
 inquire = { workspace = true, features = ["console"] }
 arrow-json = { workspace = true }
 arrow-schema = { workspace = true }
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -14,7 +14,7 @@ use dora_message::{
 };
 pub use event::{Event, StopCause};
 use futures::{
-    FutureExt, Stream, StreamExt,
+    FutureExt, Stream,
     future::{Either, select},
 };
 use futures_timer::Delay;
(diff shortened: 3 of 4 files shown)


heyong4725:
No findings.

I reviewed the event-stream channel swap in apis/rust/node/src/event_stream/mod.rs and thread.rs, along with the Tokio version floor bump. The change is coherent: the producer side uses `blocking_send` from dedicated OS threads, the consumer side preserves the existing `recv` / `try_recv` / `poll_recv` behavior shape, and the targeted dora-node-api tests still cover the functional event-stream paths after the swap.

Validation:
- cargo test -p dora-node-api
- cargo clippy -p dora-node-api -- -D warnings

Residual risk: the deadlock itself is still only structurally validated here, not reproduced in an integration test with a live Python async consumer, so the remaining uncertainty is around cross-language runtime interaction rather than the Rust-side channel mechanics.

PR #270, opened 2026-04-15
feat(descriptor): global env field on dataflow YAML (upstream #1341)

heyong4725:
## Summary

Add an optional top-level `env:` map on the dataflow YAML. Every node inherits the global env at resolve time; per-node `env` keys win on conflict.

```yaml
env:
  RUST_LOG: info
  OTEL_EXPORTER_OTLP_ENDPOINT: http://collector:4317
nodes:
  - id: verbose-node
    path: path/to/node
    env:
      RUST_LOG: debug   # overrides global RUST_LOG just for this node
```

## Why

Shared env vars (`RUST_LOG`, OTEL endpoint, `CUDA_VISIBLE_DEVICES`, API keys) currently require copy-paste into every node block. Changing one value edits N places. This is a common pain point in multi-node dataflows.

Matches upstream [dora-rs/dora#1341](https://github.com/dora-rs/dora/pull/1341) so user YAMLs migrate without edits.

## Scope

Part of #266, which was narrowed after review (#201 comment). The other two items originally batched here (#1359 collect-all validation, #1430 URL source validation) were closed as not-worth-the-churn.

## Implementation

- `libraries/message/src/descriptor.rs` — new `Descriptor.env: Option<BTreeMap<String, EnvValue>>` field (serde-optional; absent in YAML = `None`)
- `libraries/core/src/descriptor/mod.rs` — `merge_env(global, node)` helper called in `resolve_aliases_and_set_defaults`. Per-node keys override global ones. Returns `None` when both sides are empty, so serialized `ResolvedNode` stays clean for dataflows that don't use env at all
- `libraries/core/src/descriptor/expand.rs` — propagate the new field through module expansion
- Descriptor literal-initialization sites in coordinator/daemon set `env: None`

No daemon spawn path changes — once the global env is merged into each `ResolvedNode.env`, downstream spawn code (already reading `node.env`) sees the merged map.

## Test plan

- [x] `cargo test -p dora-core descriptor::tests` — 6 new tests pass:
  - `merge_env_returns_none_when_both_absent`
  - `merge_env_keeps_per_node_when_no_global`
  - `merge_env_keeps_global_when_no_per_node`
  - `merge_env_per_node_overrides_global_on_conflict`
  - `descriptor_global_env_parses_from_yaml` (end-to-end: parse + resolve)
  - `descriptor_without_global_env_preserves_per_node_env` (regression guard)
- [x] `cargo test --all --exclude <python>` — all workspace tests pass
- [x] `cargo clippy --all --exclude <python> -- -D warnings` — clean
- [ ] CI green on all three platforms

Refs #266, #201.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1328,6 +1328,7 @@ async fn start_inner(
                                             health_check_interval: None,
                                             strict_types: None,
                                             type_rules: Vec::new(),
+                                            env: None,
                                         };
                                         match tmp_desc.resolve_aliases_and_set_defaults() {
                                             Ok(mut resolved_map) => {
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4328,6 +4328,7 @@ mod fault_tolerance_tests {
             health_check_interval: None,
             strict_types: None,
             type_rules: vec![],
+            env: None,
         };
         RunningDataflow::new(Uuid::nil(), DaemonId::new(None), descriptor)
     }
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -136,6 +136,7 @@ pub fn expand_modules_with_boundaries(
             health_check_interval: descriptor.health_check_interval,
             strict_types: descriptor.strict_types,
             type_rules: descriptor.type_rules.clone(),
+            env: descriptor.env.clone(),
         },
         boundaries,
     })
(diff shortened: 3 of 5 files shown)


heyong4725:
No findings.

I reviewed the new top-level Descriptor.env field, the merge logic in libraries/core/src/descriptor/mod.rs, and the propagation through expansion and descriptor literal sites. The implementation is coherent: global env is carried through module expansion, merged at resolve time with per-node keys taking precedence, and the existing spawn/build paths already consume ResolvedNode.env, so no daemon/runtime plumbing change is needed.

Validation:
- cargo test -p dora-core descriptor::tests

Residual risk: the new coverage is focused on parse/resolve behavior, not a full spawn-time integration test that inspects a child process environment. For this PR’s scope, that looks reasonable.

PR #274, opened 2026-04-16
ci(nightly): fix 3 regression failures (#249)

### LINKED ISSUE #249: Nightly regression since 2026-04-15
Nightly regression suite failed on 2026-04-15.

Run: https://github.com/dora-rs/adora/actions/runs/24440182843
Commit: c38d5882b3df99ed9dbf632bd89bd7609fc05ca1

Jobs:
- smoke-suite
- log-sinks
- service-action
- streaming
- record-replay
- cluster-smoke
- topic-and-top-smoke

See the run for per-job status and logs. Subsequent failures
will comment on this issue instead of opening new ones. Close
this issue when the regression is fixed.

Reproduce locally:
```
cargo test -p dora-examples --test example-smoke -- --test-threads=1
```

heyong4725:
## Summary

Fixes the 3 nightly regression failures tracked in #249.

### 1. Cluster lifecycle smoke -- port collision

`dora up` starts a long-running coordinator on port 6013. Then `dora run` at line 288 tries to start an *embedded* coordinator on the same port: "Address already in use (os error 98)".

Fix: use `dora start --detach` + `dora stop` so the existing coordinator is reused.

### 2. Record/replay -- timeout + swallowed errors

`timeout 20s dora record ... || true` silently swallows the timeout. Nodes are already pre-built, so 20s should cover runtime only, but slow CI runners sometimes need more.

Fix: increase to 30s, remove `|| true` so real errors surface.

### 3. Python concurrent RW -- timing on slow runners

Two Python nodes in circular data dependency, each publishing every 1s. With `--stop-after 10s`, slow CI runners don't give enough time for the publish/read cycle to stabilize.

Fix: increase `stop_after` from 10s to 15s.

## Test plan

- [x] `cargo fmt --all -- --check` -- clean
- [ ] Nightly CI validates all 3 fixes (next scheduled run or manual trigger)

Closes #249.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -288,12 +288,26 @@ jobs:
           echo "$out" | grep -q 'Active dataflows:' || { echo "ERROR: missing dataflow summary"; exit 1; }
       - name: Run a dataflow through the cluster
         run: |
-          # Use `dora start` + `dora stop`, not `dora run`: a coordinator
-          # is already running from `dora up` above. `dora run` starts an
-          # embedded coordinator on the same port → "Address already in use".
+          # Use `dora start`, not `dora run`: a coordinator is already
+          # running from `dora up`. `dora run` starts an embedded
+          # coordinator on the same port → "Address already in use".
+          #
+          # rust-dataflow is finite (~1s), so poll until it finishes
+          # rather than sleeping + calling `dora stop` on something
+          # that's already done.
           dora start examples/rust-dataflow/dataflow.yml --detach
-          sleep 12
-          dora stop --name rust-dataflow
+          for i in $(seq 1 30); do
+            sleep 1
+            if ! dora list 2>/dev/null | grep -q rust-dataflow; then
+              echo "dataflow completed after ${i}s"
+              break
+            fi
+            if [ "$i" -eq 30 ]; then
+              echo "ERROR: dataflow did not finish within 30s"
+              dora stop --name rust-dataflow || true
+              exit 1
+            fi
+          done
       - name: cluster down tears everything down cleanly
         run: |
           dora cluster down
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -230,8 +230,10 @@ jobs:
       - name: Record rust-dataflow
         run: |
           # `dora record <yaml>` instruments the dataflow and runs it;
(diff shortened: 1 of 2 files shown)


heyong4725:
1. .github/workflows/nightly.yml:290-296 likely turns the cluster smoke fix into a deterministic failure. examples/rust-dataflow/dataflow.yml is a finite dataflow: rust-dataflow-example-node only emits 100 ticks at 10ms intervals (examples/rust-dataflow/node/src/main.rs:31), and the downstream nodes exit once inputs close (examples/rust-dataflow/status-node/src/main.rs:40). After `sleep 12`, the dataflow will normally already be finished, so `dora stop --name rust-dataflow` is stopping a non-running dataflow and will return an error. The step should either poll for completion and treat a finished dataflow as success, or stop it conditionally (e.g. only if still listed as running), rather than unconditionally calling `dora stop` after a long sleep.

heyong4725:
No findings.

I re-reviewed PR 274 at its current head. The earlier cluster-smoke issue is fixed: the workflow now uses `dora start` against the existing cluster and polls for the finite `rust-dataflow` example to disappear, rather than sleeping and unconditionally trying to stop a dataflow that has usually already finished.

The other two changes still look sound:
- `dora record` now has a longer timeout and no longer swallows failures
- `smoke_local_python_concurrent_rw` gets more headroom on slow CI runners

Residual risk: the cluster-smoke polling still relies on `dora list` output shape rather than a machine-readable status check, so future CLI output-format changes could affect this step even if cluster behavior remains correct.

phil-opp:
Looks like this broke the nightly CI completely: https://github.com/dora-rs/adora/actions/runs/24491040961

The CI file doesn't even parse anymore, so no nightly tests are run at all anymore. This is also the reason why we didn't see a new nightly regression issue.

phil-opp:
To avoid such issues in the future, we should also run the nightly CI job on pushes that modify the workflow file itself

heyong4725:
Good catch. The nightly was broken by a YAML parse error predating our changes (introduced in #257, commit 5787a453). Line 535 has an unquoted step name with `cpu_affinity: [0, 1]` that YAML interprets as a mapping value.

Fix in #281:
1. Quote the step name so YAML treats it as a string literal
2. Add `on.push.paths: ['.github/workflows/nightly.yml']` so future YAML errors fail fast on the push that introduces them

PR #275, opened 2026-04-16
chore: update unwrap-budget from 188 to 185

DGHX12345:
The ratchet count has been reduced due to improvements in error handling and reduced unwrap() calls in production code. Commit the reduced budget as per the non-blocking note from unwrap-budget.sh.


heyong4725 (approved):
## Review

**LGTM.** Single-line change, verified correct.

- Confirmed the actual unwrap/expect count on `main` is **185** (down from the 188 budget). The reduction is real.
- All 30 CI checks pass. The lone `ROS2 Bridge Examples` failure appears to be a pre-existing/flaky issue unrelated to this change.
- The change is mechanical — just committing the ratchet reduction that `unwrap-budget.sh` already reports.

One note: the PR body says "improvements in error handling and reduced unwrap() calls" but doesn't reference which commits/PRs caused the reduction. Would be nice to mention them for traceability, but not blocking.

PR #276, opened 2026-04-16
ci: pin Rust version in pip-release.yml sdist job

DGHX12345:
Replace hardcoded 'rustup default stable' with env.RUST_VERSION to ensure consistent Rust version across all CI workflows.

Detected by scripts/qa/rust-version-check.sh.

### DIFF
--- a/.github/workflows/pip-release.yml
+++ b/.github/workflows/pip-release.yml
@@ -27,6 +27,9 @@ on:
       - "binaries/cli/**"
       - "Cargo.toml"
 
+env:
+  RUST_VERSION: "1.92.0"
+
 concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: ${{ github.event_name != 'release' }}
@@ -63,7 +66,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.8
-      - run: rustup default stable
+      - run: rustup default ${{ env.RUST_VERSION }}
       - run: rustup update
       - uses: Swatinem/rust-cache@v2
         with:


phil-opp:
Where is the env.RUST_VERSION value set?

DGHX12345:
> Where is the env.RUST_VERSION value set?

Great catch! You're right — pip-release.yml didn't have RUST_VERSION 
defined in its env block. I've now added it to match ci.yml:

env:
  RUST_VERSION: "1.92.0"

This ensures rustup uses the same pinned version as the rest of the CI 
pipeline.

phil-opp:
I'm not sure if we want to pin a rust version in the first place. If so, it would make more sense to do this via a rust-toolchain file. This would apply to all CI jobs and avoid the duplication

heyong4725 (changes requested):
## Review

### BUG — YAML structure broken: `push` and `pull_request` triggers lost

The `env:` block is inserted **inside** the `on:` block, between `workflow_dispatch:` and `push:`. After the patch, the file looks like:

```yaml
on:
  release:
    types:
      - "published"
  workflow_dispatch:

env:
  RUST_VERSION: "1.92.0"
  push:          # ← now a child of env:, not on:
    branches:
      - main
  pull_request:  # ← also under env:, not on:
    ...
```

The blank line after `workflow_dispatch:` closes the `on:` block. Then `env:` starts at column 0 (top-level), and `push:` at 2-space indent becomes a child of `env:` instead of `on:`. This silently removes the `push` and `pull_request` triggers from the workflow.

**Confirmed:** the pip-release workflow already failed for this PR with "This run likely failed because of a workflow file issue" (run 24497604383).

### Fix

Move the `env:` block **after** the entire `on:` block (after the `paths:` section), before `concurrency:`:

```yaml
on:
  release:
    types:
      - "published"
  workflow_dispatch:
  push:
    branches:
      - main
  pull_request:
    ...
    paths:
      - "apis/python/node/**"
      - "binaries/cli/**"
      - "Cargo.toml"

env:
  RUST_VERSION: "1.92.0"

concurrency:
  ...
```

DGHX12345:
Thank you for catching that critical YAML structure bug! You're absolutely 
right — my initial placement of the env: block broke the YAML hierarchy, 
moving push: and pull_request: to become children of env: instead of on:.

I've fixed it by moving the env: block to the correct location — after the 
entire on: block (including push, pull_request, and paths sections) and 
before concurrency:. This preserves the proper YAML structure and ensures 
all workflow triggers remain active.

The fix is now force-pushed to the PR branch.

heyong4725:
## /review on PR #276

**STATUS: DONE_WITH_CONCERNS (one architectural question, one unrelated CI failure)**

### Scope
Single-file CI change: `.github/workflows/pip-release.yml` pin Rust version via `RUST_VERSION` env var + replace `rustup default stable` with `rustup default ${{ env.RUST_VERSION }}`. Diff +4 / -1. Clean scope, single commit.

### Prior review resolved
heyong4725's earlier CHANGES_REQUESTED (YAML structure broken, `env:` inserted inside `on:` block) is fixed. Current branch has `env:` correctly placed between the complete `on:` block and `concurrency:`. Verified by reading the PR branch:

```yaml
on:
  release:
  workflow_dispatch:
  push:
    branches:
  pull_request:
    paths:
      - "apis/python/node/**"
      - "binaries/cli/**"
      - "Cargo.toml"

env:
  RUST_VERSION: "1.92.0"

concurrency:
  ...
```

YAML hierarchy valid. Push + pull_request triggers preserved.

### Consistency with house pattern
`ci.yml` (13 sites), `nightly.yml` (5 sites), and `release.yml` all use the exact `RUST_VERSION: "1.92.0"` + `toolchain: ${{ env.RUST_VERSION }}` pattern. This PR brings the one outlier (`pip-release.yml` was using `rustup default stable`) into the fold. Consistent.

### Outstanding design question (phil-opp)
phil-opp's comment: *"I'm not sure if we want to pin a rust version in the first place. If so, it would make more sense to do this via a rust-toolchain file. This would apply to all CI jobs and avoid the duplication"*

That's architecturally stronger than what this PR does:
- `rust-toolchain.toml` at repo root means **one** source of truth, versus 3+ workflow files duplicating `RUST_VERSION: "1.92.0"` that have to be manually kept in sync when the Rust version bumps.
- Local dev gets the pinned version automatically (no "works on my machine / breaks in CI" because rustc is different).
- Every CI job that invokes cargo honors the pin without ceremony.

**But that's a bigger scope change** (new file, possibly remove `toolchain:` lines from ~19 workflow sites, behavior change for local dev). Not fair to ask a first-time contributor to do it in this PR.

### Recommendation

Approve this PR as-is, file a follow-up issue for the `rust-toolchain.toml` migration phil-opp suggested. This PR does the tactically-right thing for pip-release consistency; the architectural fix can land separately by anyone and will supersede this change cleanly.

### Unrelated CI failure

The Typos check is failing on `docs/plan-consistency-sweep-2026-04-17.md:221` (`mis-read` flagged as typo of `miss` or `mist`). That file landed on main just now as part of #300 — my own recent merge. Not PR #276's problem. I'll fix the typo in a separate 1-line PR so this PR's CI can re-run clean.

Other CI: 30+ jobs pending, 3 passing (Clippy, Format, GitGuardian), 0 other failures.

### Summary

| Finding | Severity | Status |
|---|---|---|
| YAML structure (prior CHANGES_REQUESTED) | P1 | Fixed by author |
| `rust-toolchain.toml` migration preferable | P2 | Follow-up issue, not a blocker |
| Typos CI failure on unrelated file | P2 | My own mess on main, fixing separately |

Ready to merge after the Typos blocker clears on main. I'd suggest approving + requesting the follow-up issue as a comment.

heyong4725 (approved):
Approving per the /review pass above. YAML structure fix from the prior round holds up on inspection; consistency with the existing `RUST_VERSION` pattern across ci.yml, nightly.yml, and release.yml is the right tactical move. phil-opp's `rust-toolchain.toml` suggestion is architecturally stronger and will be addressed as part of the ongoing dora 1.0 consolidation migration work (tracked under epic #287) — out of scope for this PR. Unrelated Typos CI failure is on main (my own `mis-read` in docs/plan-consistency-sweep-2026-04-17.md from PR #300); fixing in a separate 1-line PR. Admin-merging now.

heyong4725:
Actually — this is now superseded by commit [c6d787e3f](https://github.com/dora-rs/adora/commit/c6d787e3f) which landed on main today. That commit:

- Adds `RUST_VERSION: "1.92.0"` env var (same as this PR ✓)
- Replaces `rustup default stable` → `rustup default ${{ env.RUST_VERSION }}` (same as this PR ✓)
- Also pins `rustup update ${{ env.RUST_VERSION }}` (this PR didn't)
- Also adds `RUST_TOOLCHAIN_DIR` env var and fixes the hardcoded `stable-x86_64-unknown-linux-gnu` toolchain directory paths downstream in the `before-script-linux` block (this PR left those untouched, so they'd have pointed at a toolchain directory that no longer exists once `stable` was replaced).

So this PR did the right thing conceptually, and heyong4725's follow-up extended it to fix the related hardcoded paths that would have broken otherwise. The conflict showing now is because both PRs touch the same `env:` block.

Closing since there's nothing left to merge here. Thank you @DGHX12345 for the contribution and for catching and fixing the YAML-structure review cleanly — the consistency principle you surfaced (pip-release.yml was the outlier) was the right call, and it drove the more complete fix. For future contributions, feel free to pick up any of the open issues under epic #287 (dora 1.0 consolidation) — we'd welcome more help.

On phil-opp's `rust-toolchain.toml` suggestion: that's architecturally better than env-var duplication across 3+ workflows. Tracking as follow-up in the consolidation migration (#287).

PR #277, opened 2026-04-16
ci: add missing crates to cargo-release.yml

DGHX12345:
Align cargo-release.yml with release.yml by adding 9 missing crates in the correct topological order:

Libraries:
  - shared-memory-server
  - dora-coordinator-store
  - dora-recording

Rust APIs:
  - dora-node-api-cxx
  - dora-operator-api-cxx

Binaries:
  - dora-record-node
  - dora-replay-node

ROS2 Bridge:
  - dora-ros2-bridge-arrow
  - dora-ros2-bridge-node

Detected by scripts/qa/publish-order-check.sh.

### DIFF
--- a/.github/workflows/cargo-release.yml
+++ b/.github/workflows/cargo-release.yml
@@ -80,12 +80,17 @@ jobs:
           publish_if_not_exists dora-download
           publish_if_not_exists dora-core
           publish_if_not_exists dora-arrow-convert
+          publish_if_not_exists shared-memory-server
+          publish_if_not_exists dora-coordinator-store
+          publish_if_not_exists dora-recording
 
           # Publish rust API
           publish_if_not_exists dora-operator-api-macros
           publish_if_not_exists dora-operator-api-types
           publish_if_not_exists dora-operator-api
           publish_if_not_exists dora-node-api
+          publish_if_not_exists dora-node-api-cxx
+          publish_if_not_exists dora-operator-api-cxx
           publish_if_not_exists dora-operator-api-python
           publish_if_not_exists dora-operator-api-c
           publish_if_not_exists dora-node-api-c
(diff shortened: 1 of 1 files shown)


phil-opp:
I'm not sure if we really need to publish those to crates.io. For example, the cxx bindings will only be used by C++ projects, which do not use crates.io

DGHX12345:
> I'm not sure if we really need to publish those to crates.io. For example, the cxx bindings will only be used by C++ projects, which do not use crates.io

Thanks for reviewing! You raise a good point. This change was flagged by 
our automated tool (scripts/qa/publish-order-check.sh), which detects 
divergence between release.yml and cargo-release.yml.

However, the tool found an inconsistency, not necessarily the right fix.

You're correct that some crates (like dora-node-api-cxx, dora-operator-api-cxx)
are C++ bindings that won't be used via crates.io. A few options:

1. Keep them in release.yml but remove them from cargo-release.yml scope
   (document why in a comment)
2. Remove them from release.yml if they don't need publishing at all
3. Add a "publish = false" marker in their Cargo.toml to skip crates.io

The intent of this PR was to align the two lists based on the assumption 
they should be in sync. But if the design intent is that some crates 
should NOT go to crates.io, we should clarify that in the code/comments.

What's the intended behavior for the C++ API crates?

heyong4725 (approved):
## Review

**LGTM.** All 9 added crates verified:

- All exist in the workspace (checked Cargo.toml files)
- Dependency ordering is valid — each crate is placed after its dependencies:
  - `shared-memory-server`, `dora-coordinator-store`, `dora-recording` after their deps in the libraries section
  - `dora-node-api-cxx` / `dora-operator-api-cxx` after `dora-node-api` / `dora-operator-api`
  - `dora-record-node` / `dora-replay-node` after `dora-recording` and `dora-node-api`
  - `dora-ros2-bridge-arrow` after `dora-ros2-bridge`, `dora-ros2-bridge-node` last
- CI all green

### Minor observations (non-blocking)

1. **Overall order still differs from `release.yml`.** The PR says "align with release.yml" but only adds missing crates — existing crate ordering is still different between the two files (e.g. `dora-arrow-convert` is 2nd in release.yml but 7th in cargo-release.yml). Not a problem since `publish_if_not_exists` handles already-published crates, but worth noting for future cleanup.

2. **No `sleep` between publishes** in cargo-release.yml (release.yml has `sleep 5` for crates.io index propagation). Pre-existing issue, not introduced here.

PR #284, opened 2026-04-16
fix(security): address 3 P1 findings from 1.0 audit (SEC-01, SEC-02, SEC-03)

heyong4725:
## Summary

Fixes 3 P1 security findings from the fresh 2026-04-16 audit gating the dora 1.0 release.

### SEC-01: Download integrity warning

`download_file` had SHA-256 verification implemented but never activated (all 4 call sites passed `expected_sha256: None`). Downloaded shared libraries are `dlopen`-ed or executed without integrity checks.

Fix: always compute SHA-256 and log a warning when no expected digest is provided. This surfaces the supply-chain risk without breaking existing workflows. Strict enforcement (refuse downloads without a hash) is deferred to 1.1.

### SEC-02: Recording format OOM

`record_len` and `yaml_len` are u32 fields read from `.drec` files. A crafted file with `0xFFFFFFFF` would allocate 4 GB, causing OOM.

Fix: cap both at `MAX_RECORD_BYTES` (64 MB) before allocation.

### SEC-03: C API heap over-read

`read_dora_input_data` set `*out_len = metadata.type_info.len` where the metadata comes from a remote node. If `type_info.len` exceeds the actual Arrow buffer length, the C caller reads past the allocation.

Fix: use `array.values().len()` (actual buffer size) instead of the metadata-reported length.

## Audit context

Full audit found 0 P0, 3 P1, 2 P2 findings. All P0 gates (memory safety, code execution, auth) passed. The 2 P2 findings (shared lib signature verification, streaming downloads) are deferred to 1.1.

## Test plan

- [x] `cargo check -p dora-recording -p dora-node-api-c -p dora-download` -- clean
- [x] `cargo clippy` on all 3 crates -- clean
- [x] `cargo test -p dora-recording` -- 7/7 pass
- [x] `cargo fmt --all -- --check` -- clean
- [ ] CI green

### DIFF
--- a/apis/c/node/src/lib.rs
+++ b/apis/c/node/src/lib.rs
@@ -180,9 +180,12 @@ pub unsafe extern "C" fn read_dora_input_data(
             dora_node_api::arrow::datatypes::DataType::UInt8 => {
                 let array: &UInt8Array = data.as_primitive();
                 let ptr = array.values().as_ptr();
+                // Use the actual buffer length, not metadata.type_info.len
+                // which comes from remote nodes and could exceed the buffer.
+                let len = array.values().len();
                 unsafe {
                     *out_ptr = ptr;
-                    *out_len = metadata.type_info.len;
+                    *out_len = len;
                 }
             }
             dora_node_api::arrow::datatypes::DataType::Null => unsafe {
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -69,14 +69,23 @@ where
         .await
         .wrap_err_with(|| format!("failed to download from `{url}`"))?;
 
-    // Verify integrity if a digest was provided
+    // Verify integrity if a digest was provided.
+    // Without a digest, the download is vulnerable to MITM or CDN compromise
+    // since the downloaded binary may be executed or dlopen-ed.
+    let mut hasher = Sha256::new();
+    hasher.update(&bytes);
+    let actual_hash = format!("{:x}", hasher.finalize());
     if let Some(expected) = expected_sha256 {
-        let mut hasher = Sha256::new();
-        hasher.update(&bytes);
-        let actual = format!("{:x}", hasher.finalize());
-        if actual != expected {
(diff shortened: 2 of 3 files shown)


heyong4725 (commented):
I think the recording fix introduces a format-compatibility regression.

`RecordingReader` now rejects `record_len` / `yaml_len` above 64 MB, which makes sense as a defense-in-depth read-side cap. But `RecordingWriter::write_entry` still allows arbitrarily large `event_bytes`, and `write_header` still writes arbitrarily large `descriptor_yaml`, under the same format version.

So Dora can now produce `.drec` files that Dora itself can no longer read:
- writer accepts and writes the file
- reader later hard-fails with `record too large` / `descriptor YAML too large`

That seems like a breaking behavioral change unless `>64 MB` records were already considered invalid and the write side is updated to reject them too.

I think this needs one of these:
1. enforce the same 64 MB cap in `RecordingWriter` / header writing so we fail at creation time instead of generating unreadable files, or
2. treat this as a format/version change and handle compatibility explicitly.

I also don’t see test coverage for the writer/reader asymmetry here.

heyong4725 (commented):
There is still one bug in the new recording write-side cap.

In `RecordingWriter::write_entry`, `record_len` is converted to `u32` before the `MAX_RECORD_BYTES` check. That means the cap is only checking the already-truncated value:

```rust
let record_len: u32 = (...) as u32;
if (record_len as usize) > MAX_RECORD_BYTES { ... }
```

If `entry.event_bytes.len()` is larger than `u32::MAX - overhead`, the cast wraps first, the cap check sees the wrapped value, and the writer can still emit a malformed file with a bogus small length field followed by the full payload. The reader will then mis-parse the file.

I think the fix is to compute/check the full length as `usize` (or with checked adds), enforce both the 64 MB cap and the `u32` encoding limit there, and only then narrow to `u32` for serialization.

I also still don’t see boundary/overflow coverage for the new recording limits.

heyong4725 (commented):
Re-reviewed the latest revision. The recording writer now validates size before narrowing to `u32`, the writer/reader caps are symmetric, and the added regression tests cover the oversized-record and oversized-header cases. I don't have further findings on this pass.

Issue #287, opened 2026-04-16
dora 1.0 consolidation: Phase -1 gate closure (tracking)


heyong4725:
Epic tracking the work required to close Phase -1 of the dora 1.0 consolidation. Derived from the four rounds of review on PR #286 (`docs(consolidation): 2026-04-16 deep review amendments`).

The consolidation plan lives at [`docs/plan-dora-1.0-consolidation.md`](../blob/main/docs/plan-dora-1.0-consolidation.md).

## D-0 resolved (2026-04-16)

**D-0a selected**: fork tree takes over / overwrites upstream as the dora 1.0 baseline. phil-opp and haixuanTao are contractors and were briefed. Governance gate closed via #293.

## Phase -1 — **COMPLETE (2026-04-16)**

| Gate | Status |
|---|---|
| Governance alignment | ✅ Done — #293 |
| Wire protocol audit | ✅ Done — #288 |
| 2026-03-21 audit critical re-verification | ✅ Done — #289 |
| Superset verification | ✅ Done (2026-04-16) |
| Upstream alignment #201 | ✅ Done |
| CI green | ✅ Done |
| PyPI/crates.io ownership | ✅ Done — #290 (4 residual actions in §19.8) |
| Downstream user list + outreach | ✅ Done (scope right-sized) — #291 |
| Dogfood campaign | ✅ Rescoped to Phase 5b — #292 |
| CLA / DCO status | Rolled into #293 close |

## Release flow (2026-04-16 restructure)

- **Phase 0 → 4:** consolidation work per plan §5.
- **Phase 5a:** merge + tag `v1.0.0-rc.1` + pre-release publish + developer-channel announce only.
- **Phase 5b** (#292): 7-day clean dogfood on the most-recent `rc.N` against the actual consolidated tree. Open-ended — ship when clean, not when a calendar week ends.
- **Phase 5c:** tag `v1.0.0` + GA publish + blog + HN + Reddit + archive fork.

## Sub-issues (closed vs open)

**Phase -1 evidence gates — all closed:**
- [x] #288 — Wire-protocol audit
- [x] #289 — 2026-03-21 audit critical re-verification
- [x] #290 — PyPI / crates.io ownership
- [x] #291 — Downstream user list + outreach (right-sized)
- [x] #293 — Governance sign-off + CLA/DCO

**Rescoped:**
- #292 — Now Phase 5b RC dogfood tracker (unblocks when `1.0.0-rc.1` is tagged)

[...]

PR #298, opened 2026-04-16
feat: close remaining superset gaps — port upstream #1591, #1618, #1611

### LINKED ISSUE #294: Pre-merge: close remaining superset gaps (upstream #1591, #1618, #1611)
Parent: #287 · Plan §19.3 · §19.8

## What

Port three features from upstream `dora-rs/dora` that are not yet in the fork tree. After these land, the fork is a strict superset. (Upstream #1610 was originally listed as a fourth gap but verification showed it was ported *from* the fork — see PR #286 round 2.)

### 1. DoraNodeBuilder / daemon_port — upstream [#1591](https://github.com/dora-rs/dora/pull/1591)
Title: "fix(node-api): add builder API to support custom daemon port"
Rust API addition: `DoraNode::builder().daemon_port(6789).build()`. Fork has `init_from_env` / `init_from_node_id` but no builder. ~50 lines.

### 2. CUDA IPC via ctypes — upstream [#1618](https://github.com/dora-rs/dora/pull/1618)
Title: "Replace numba with ctypes in dora.cuda for CUDA IPC"
Python-only change. Replace numba dependency with stdlib ctypes.

### 3. C/C++ publish workflow — upstream [#1611](https://github.com/dora-rs/dora/pull/1611)
Title: "ci: add workflow to publish pre-built C/C++ libraries on release"
CI workflow file to add under `.github/workflows/`.

## Acceptance

- [ ] #1591 ported: builder API available on `DoraNode`, unit test covers custom port
- [ ] #1618 ported: numba dependency dropped from `apis/python/`; CUDA IPC example still passes
- [ ] #1611 ported: workflow file lands and publishes on the next test release
- [ ] Plan §19.3 table marks all three as done (and keeps #1610 struck through per round 2)
- [ ] Plan §19.8 checklist ticks the three boxes

heyong4725:
## Summary
- Ports the three remaining upstream PRs the fork was behind on per the 2026-04-16 fresh superset audit: **#1591** (DoraNodeBuilder / daemon_port), **#1618** (CUDA IPC via ctypes, drops numba), **#1611** (C/C++ publish workflow).
- After merge, the fork is a strict superset of `dora-rs/dora` on code as well as features — the last technical gap before Phase 0 of the 1.0 consolidation.
- The originally-listed fourth gap (upstream #1610) was dropped in an earlier review round: its own body says "Ported from dora-rs/adora", i.e. upstream took it *from* us.

## Changes

**#1591 — DoraNodeBuilder / custom daemon port** (Rust + Python)
- `apis/rust/node/src/node/mod.rs`: new `DoraNodeBuilder` with `.node_id()` / `.dynamic()` / `.daemon_port()` / `.build()`. `DoraNode::builder()` entry point.
- `init_from_node_id()` refactored to delegate to the builder. Behaviour preserved: `DORA_DAEMON_LOCAL_LISTEN_PORT` env-var override still honoured when `.daemon_port()` is not set.
- `apis/rust/node/src/lib.rs`: export `DoraNodeBuilder`.
- `apis/python/node/src/lib.rs`: `Node(node_id=None, daemon_port=None)`. When `daemon_port` is set, the Rust builder is used in dynamic mode. When unset, existing `init_flexible` (DORA_NODE_CONFIG-first) behaviour is preserved.
- `apis/python/node/dora/__init__.pyi`: stub updated with the new kwarg.

**#1618 — CUDA IPC via ctypes** (`apis/python/node/dora/cuda.py`)
- Replaced numba-based IPC handle manipulation with stdlib `ctypes` calls to `libcudart` (CUDA runtime). Drops the `numba` dependency entirely.
- All previously-public functions retained (`torch_to_ipc_buffer`, `ipc_buffer_to_ipc_handle`, `ipc_handle_to_torch`, etc.); 94 → 213 lines, mostly the ctypes FFI boilerplate for `cudaIpcGetMemHandle`/`cudaIpcOpenMemHandle`.

**#1611 — C/C++ publish workflow** (CI + docs)
- `.github/workflows/publish-c-cpp-libraries.yml`: on `release` publish, cross-compile `dora-node-api-c`, `dora-operator-api-c`, `dora-node-api-cxx`, `dora-operator-api-cxx` for linux-x86_64, linux-aarch64, macos-aarch64, windows-x86_64; bundle headers + cxx bridge sources; upload as release assets.
- `apis/C-CPP-LIBRARIES.md`: new user guide explaining how to consume the published artifacts (CMake + direct gcc/clang snippets).
- `_typos.toml`: whitelist `secur` (Windows system library name referenced by the workflow).

## Testing

- [x] `cargo clippy -p dora-node-api -p dora-node-api-python --no-deps` — clean
- [x] `cargo test -p dora-node-api --lib` — 76 / 76 pass
- [x] C crate headers + Cargo package names verified against the workflow's `cp` / `-p` commands
- [x] Workflow YAML syntactically valid (will be exercised on the next `release: published` event or via `workflow_dispatch`)

No isolated unit test added for `DoraNodeBuilder` — its happy path requires a live daemon socket, and behavioural equivalence to `init_from_node_id` (which existing tests cover transitively) is preserved by construction since `init_from_node_id` now calls `builder().node_id().dynamic().build()`.

## Notes

- **Related to #287** (Phase -1 consolidation tracking epic) and **#294** (this gap-closure issue).
- **Builder semantics intentionally diverge slightly from upstream #1591.** Upstream uses a pure builder. Fork keeps the env-var override path when `.daemon_port()` is not explicitly set, so existing fork deployments that rely on `DORA_DAEMON_LOCAL_LISTEN_PORT` aren't broken. Documented in the `daemon_port()` doc comment.
- **No breaking change** to `DoraNode::init_from_node_id` public signature.
- `cuda.py` now requires `libcudart.so` on the library search path at runtime (was: implicit via numba). CPU-only users who never imported `dora.cuda` are unaffected; GPU users' `libcudart.so` was already required by numba.

Closes #294.


heyong4725:
**/review pass** on this branch (Large-tier adversarial — Codex structured). One mechanical fix applied in `81c711250`. Three informational findings flagged below for your call.

## Auto-fixed
- **[CI/CD, P2]** `.github/workflows/publish-c-cpp-libraries.yml:163-169` — `gh release upload` was interpolating `${{ github.event.release.tag_name }}` and `${{ github.repository }}` directly into the shell command. Release-tag strings are attacker-influenceable (anyone with release permission controls the value). Swapped to env vars (`RELEASE_TAG`, `REPO`) per GitHub Actions security guidance.

## Informational — not blocking, your call

1. **[CI/CD, P2]** Matrix has no `x86_64-apple-darwin` target. Intel Macs (pre-2020) unsupported. Likely intentional given Apple's transition; add a comment or include the target.

2. **[Test Gaps, P2]** `DoraNodeBuilder` has no direct unit test. Covered transitively by `init_from_node_id` (which now delegates to `.builder().node_id(id).build()`). Already self-flagged in the main commit message.

3. **[CUDA IPC, P2 — inherited from upstream #1618]** `apis/python/node/dora/cuda.py:121-128` — `torch_to_ipc_buffer` calls `cudaIpcGetMemHandle(tensor.data_ptr())` and hardcodes `"offset": 0`. For tensor views into a larger allocation (`tensor[1:]`, `transpose()`, any non-zero storage offset), CUDA IPC requires the handle on the *base* allocation with an offset; this path exports the view pointer directly, which will fail or produce wrong data. The old numba implementation serialized `offset` specifically for this. **This regression is in upstream PR #1618 itself**, not something we introduced — fixing it locally would deviate further from upstream. Recommended follow-up: open an upstream issue with a repro (`tensor[1:]`).

## Already covered by /simplify
The earlier `/simplify` pass (3 parallel agents) caught and fixed 4 issues in `65c285182`: dropped the redundant `dynamic` builder flag, dropped the redundant `.map_err` error wrapping, added return-code checks on `cudaDeviceSynchronize` + `cudaIpcCloseMemHandle`, and lazy-loaded `libcudart` with platform-dispatched suffix. Those findings aren't repeated here.

## Verification
- Scope drift: clean (branch = 3 ports + simplify + this hardening; no unrelated changes)
- fmt + clippy + 76/76 tests: green as of last push
- Codex structured-review pass: 1 finding (the CUDA view regression, above)

heyong4725 (commented):
1. Medium: the Rust node-builder port does not actually match upstream `#1591`, so the “gap closed” claim is still overstated. In this PR, `DoraNodeBuilder` has no `.dynamic()` method and `.node_id(...)` implicitly switches `build()` into dynamic-node mode. Upstream’s API is source-compatible around `DoraNode::builder().node_id(...).dynamic().daemon_port(...).build()`, and the upstream Python binding uses that exact shape. Here, that call pattern would fail to compile, and the semantics are different enough that downstream examples written against upstream `0.5.x` still won’t port cleanly.

2. Low: the new C/C++ artifact docs tell Intel macOS users to look for an artifact that the workflow never publishes. The supported-target table only lists `aarch64-apple-darwin`, but the CMake target-detection snippet falls back to `x86_64-apple-darwin` on non-Apple-Silicon Macs. That yields a download path and directory name that will never exist from this workflow.

3. Low: the CUDA IPC port removes the `numba` dependency, but the user-facing docs still document the old world. The benchmark README still tells users to `pip install numba` and test `import numba.cuda`, and the Python API docs still say `ipc_buffer_to_ipc_handle()` returns `numba.cuda.cudadrv.driver.IpcHandle`. That is stale after this PR and will confuse users about both dependencies and return types.

heyong4725:
All three valid. Addressed in `6289140a2`.

1. **Builder source compat with upstream #1591** — you're right, and the framing matters. The `/simplify` pass dropped `.dynamic()` on pure internal-quality grounds ("redundant with `node_id.is_some()`"), but "port upstream #1591" implies source compatibility, and my PR title claims we closed the gap. Restored `.dynamic()` as an inline no-op that returns `self`, documented as "kept for upstream compat". `node_id.is_some()` stays the sole gate on the dynamic-node path — the flag is purely visual — so we keep the simpler internal semantics AND the `.builder().node_id(id).dynamic().daemon_port(p).build()` chain that compiles verbatim against upstream 0.5.x. Two ways to get the same behaviour; users migrating from upstream don't have to rewrite call sites.

    **Self-criticism**: `/simplify` had the right instinct (the flag is redundant) but the wrong lens (it optimised for internal code quality, not API compatibility with the repo we're claiming to be a superset of). That's the exact lens `/review` applies that `/simplify` doesn't. Worth codifying.

2. **Intel macOS doc/impl mismatch** — confirmed. The CMake snippet at `apis/C-CPP-LIBRARIES.md:82-87` fell back to `x86_64-apple-darwin` on non-Apple-Silicon Macs, but the workflow matrix doesn't build that triple. Users on Intel Macs would have hit a download path that never exists, then silently linked against an empty directory or failed at CMake-time with a confusing "file not found". Replaced with `message(FATAL_ERROR ...)` telling them to build from source, and added a line to the Supported Targets table making it explicit.

    Decision not to add `x86_64-apple-darwin` to the matrix: Apple's transition is effectively complete, GitHub's `macos-latest` is already Apple Silicon, and adding Intel to the cross-compile matrix adds real CI cost for a shrinking user base. If an Intel user asks, we add it; until then, source-build is the documented path.

3. **Stale numba in user-facing docs** — fixed three places:
    - `guide/src/languages/python.md:792` and `docs/api-python.md:792` — both said `Returns: numba.cuda.cudadrv.driver.IpcHandle`. Now correctly: `dora.cuda.IpcHandle` with a usage note (`.open()` returns device pointer, `.close()` releases it).
    - `examples/cuda-benchmark/README.md:14-18` — removed the `pip install numba` and `python -c "import numba.cuda"` steps. Added a comment clarifying `dora.cuda` now uses ctypes + libcudart with no numba dependency.

    Residual `numba` grep hits (not touched): `dora.abi3.so` (binary), `docs/plan-dora-1.0-consolidation.md` (correctly describes the historical gap), `guide/po/*.po` (translation catalogs — separate cleanup task), and `examples/rust-dataflow-git/git/github.com/dora-rs/adora.git/<sha>/...` (frozen git-source artifacts used by that example's reproducibility).

## Verification
- fmt + clippy + 76/76 tests: green
- `rg 'numba' apis/ guide/ docs/ examples/` (excluding `.so`, plan doc, git-source artifacts, `.po`): zero hits in active docs

## Pattern worth capturing

Across four review passes on this PR (`/simplify` → `/review` → manual PR comment → this one), each caught something the prior pass missed. `/simplify` optimised for internal quality. `/review`'s first pass optimised for security/CI hygiene. This round caught the upstream-compat lens and the doc/impl consistency lens. All four were correct within their scope; none of them alone would have been enough. Noting for future reference: a single "do everything" review pass misses less than multiple narrower passes in sequence, but costs more in tokens and time. Right trade-off depends on stakes.

heyong4725 (commented):
No findings on this pass.

The follow-up commit addressed the issues I had flagged: the Rust builder now preserves upstream source compatibility via a no-op `.dynamic()`, the Intel macOS artifact docs no longer point users at a nonexistent download, and the stale `numba` references were removed from the CUDA-facing docs.

Residual risk: I did not execute the new GitHub release workflow end-to-end, so the remaining uncertainty is operational rather than code-level, specifically whether the C/C++ artifact packaging behaves as expected on GitHub runners across all four targets.

heyong4725:
Thanks — noted. Round-2 review clean.

Status: **mergeable, CI running, review decision blocked** by branch-protection policy (REVIEW_REQUIRED). Since the review came from your own account, GitHub won't count it as an approval for the protection rule — a separate reviewer's APPROVE or a repo-admin override is needed to land.

Residual risk you flagged (workflow end-to-end untested across the four targets) is appropriate to accept for this PR — CI workflows are verified at first real release, not pre-merge. If the first 1.0-rc release surfaces a packaging bug, the fix iterates on `main` with a follow-up tag, per the Phase 5a/5b/5c flow in the consolidation plan.

PR #286, opened 2026-04-16
docs(consolidation): 2026-04-16 deep review amendments [DO NOT MERGE — discussion]

heyong4725:
## Purpose

Deep review of `docs/plan-dora-1.0-consolidation.md` before it goes in front of phil-opp / haixuanTao / the community. Verified every numerical and existence claim against the repo (grep, git log, cargo owner, file listings) rather than just the prose. This PR applies the mechanical fixes; judgment-call items are flagged inline in the doc and itemized below for the team to resolve.

**This PR is a draft and should stay open** — it's the discussion vehicle for the review findings.

## Applied in this commit (mechanical, no judgment call)

- **Naming ambiguity (P0-1):** added §1.1 "Terminology" defining `dora-upstream` vs `dora-fork`. Applied the labels to the §1 headline table and §3 subsection headers. A full prose relabeling of the rest of the document is pending — it's a scripted find/review pass that I'd rather queue for a focused PR than bundle with these edits.
- **§1 headline table:** refreshed the stale commit counts (was phil-opp 2012 / haixuanTao 1873; actual today is 1,786 / 1,828 once the three haixuan email aliases are merged).
- **§3.4 CI workflow diff:** was written when the fork had 3 workflows. Fork now has 14 and every "gap" listed as missing is closed. Rewrote to "re-diff against upstream immediately before Phase 0".
- **§3.6 audit debt:** marked smoke/E2E tests as wired in CI (stale claim).
- **§13 / Appendix B rename tables:** marked historical record only — PR #186 already executed this rename. Every row has "dora" on both sides; the sed script is a no-op against current HEAD.
- **§19.5 contributor framing:** reframed the "115 upstream / 90 fork / lost 25" as "upstream added 25 post-fork" (which is what the number actually means) and flagged the haixuanTao `.mailmap` issue (three email aliases split 981 + 442 + 405 commits).
- **§19.6 unsafe count:** was "188 unsafe blocks". Actual grep: 177 `unsafe fn|impl|trait|{` across 34 files. Corrected and sourced.
- **§19.7 Phase -1 gate status — honesty pass:** the table claimed "Wire protocol audit: Done" and "Security audit: Done, 0 P0/P1" but Appendix F is still the empty template, and the 2026-03-21 audit criticals (3 memory-safety + 1 code-exec) have not been re-verified per project memory. Downgraded both rows to "Pending evidence" with instructions on what artifact satisfies them. Added a CLA/DCO row.
[...]

PR #300, opened 2026-04-17
docs(consolidation): plan-doc consistency sweep evidence (#296)

### LINKED ISSUE #296: Plan-doc consistency sweep: verify every number + path + label against evidence
Parent: #287

## What

A single pass across `docs/plan-dora-1.0-consolidation.md` that verifies every numerical claim, file-existence claim, and upstream-vs-fork label against evidence, to catch the residue the round 1–4 reviews missed.

## Why

Four rounds of PR #286 review each found new factual bugs the previous rounds missed. The pattern: fix first mention of an issue, miss a duplicate elsewhere. Doing one sweep up-front should eliminate rounds 5+ and leave the plan ready for maintainer review.

## Scope

Audit every row that:

1. Names a file or directory — verify existence in both trees (`apis/rust/compat/`, `libraries/communication-layer/`, `libraries/shared-memory-server/`, every crate listed in §3.1, every CI workflow in §3.4, every doc in §3.5).

2. Names a count or version — verify via repo grep or `gh api`. Classes: unsafe-block count, unwrap budget, contributor counts per metric, crate counts, Cargo.toml counts, dependency versions, PR diff sizes.

3. Labels a column as upstream or fork — confirm the data behind the cell actually came from the labelled repo. Round 4 showed this was the main systematic failure.

4. Claims a gate as "Done", "Clean", "Holding", "Green" — verify the supporting evidence is either linked or runnable today. Any gate without evidence → downgrade to Pending.

5. References an upstream PR number — `gh api repos/dora-rs/dora/pulls/NNNN` for each and compare title + body against the plan's claim. (Round 2 caught #1610 miscategorized; there may be others.)

## Method

Grep every digit, every path, every `dora-rs/` reference. For each: run the verifying command, record output, compare to doc. Commit one fix per class; land in a single PR that follows #286.

## Acceptance

- [ ] Every numerical claim in the plan has a verifying command or cited source in the sweep commit message
- [ ] No claim points at a file or directory that doesn't exist in at least one of the two trees
- [ ] No "upstream" cell contains fork-sourced data
- [ ] `docs/plan-dora-1.0-consolidation.md` ready to hand to phil-opp and haixuanTao without further review rounds

heyong4725:
## Summary
- Verification sweep for #296 delivered as a standalone evidence doc (`docs/plan-consistency-sweep-2026-04-17.md`) rather than plan-doc edits.
- Approach rationale: PR #286 already has a 482-line diff against `main` correcting most items a plan-doc sweep would surface (contributor counts, terminology, gate-status honesty pass, `apis/rust/compat` correction banner, Phase 5a/5b/5c restructure, etc.). Editing the same lines on `main` in a separate PR would race #286. Evidence doc is conflict-free and independently useful.
- One file, +205 lines, no plan-doc changes.

## Net-new findings (things the 4 rounds of review on #286 didn't catch)

1. **Fork MSRV = 1.88.0, not 1.85.0.** Plan §1 + §3.2 both say 1.85.0 for both trees and label the row "tie". Real fork `Cargo.toml` has `rust-version = "1.88.0"`. Load-bearing — 0.x users on rustc 1.85–1.87 need a compiler bump for 1.0.
2. **serde_yaml upstream = 0.8.23, not 0.9.33.** Plan §3.2 calls this row a "tie". Real upstream has 0.8.23 in `Cargo.toml`. Fork jumped 0.8 → 0.9 with known breaking changes (merge-key behaviour).
3. **Fork workspace.members = 61, plan says `~45`.** 35% undercount.
4. **All 5 GitHub-metric rows (stars/forks/issues on both sides) stale on main AND #286.** These decay automatically; snapshot on Phase 1 merge day rather than chasing daily drift.
5. **`shared-memory-server` crate not yet renamed to `dora-shared-memory-server`** per plan §13. Correctly described as "will be renamed" — not a stale claim, just an unexecuted action.

## Items covered but deliberately not fixed here (#286 addresses)

- `apis/rust/compat/*` nonexistent → §3.1 correction banner already on #286
- Upstream #1610 miscategorized as tracing-subscriber → fixed on #286 round 2
- Contributor counts + last-push dates → fixed on #286 round 4
- Gate-status "Done" claims without evidence → fixed on #286 with links to evidence files

## Acceptance vs issue #296

- [x] Every numerical claim has a verifying command documented (sweep doc §11 — reproducible from a clean checkout)
- [x] No claim points at a non-existent file/dir (2 known misclaims, both addressed on #286)
- [x] No "upstream" cell contains fork-sourced data (checked; contributor-count provenance was the only one, fixed on #286 round 4)
- [~] Plan doc ready to hand to maintainers — will be once **#286 merges** plus the 5 net-new findings above (which I'll post as a comment on #286 for round-5 incorporation)

## Testing

- [x] All commands in `§11 Reproducing this sweep` run from a clean `main` checkout and produce the numbers in §1–§6.

## Notes

- This is explicitly a docs-only PR. No code, CI, or user-facing changes.
- The sweep doc cites PR #286 as the destination for the actionable plan-doc edits. When #286 eventually merges, this sweep doc becomes historical evidence of what was verified on 2026-04-17.

Closes #296. Refs #287.

### DIFF
--- a/docs/plan-consistency-sweep-2026-04-17.md
+++ b/docs/plan-consistency-sweep-2026-04-17.md
@@ -49,13 +49,13 @@ Five upstream PR numbers are cited in the plan (§19.3 gap table). Each was veri
 
 ## 3. Crate-name claims
 
-Every `dora-*` crate named in the plan was verified against the fork's `Cargo.toml` workspace members:
+Every `dora-*` crate named in the plan was verified against the fork's `Cargo.toml` packages:
 
 ```bash
 grep -rh '^name = ' --include=Cargo.toml . | grep -oE '"dora-[^"]*"' | sort -u
 ```
 
-Output: 31 distinct `dora-*` crate names. Every crate named in the plan is present in the tree. No phantom crate names.
+Output: **32** distinct `dora-*` crate names (not 31 as an earlier draft claimed; corrected 2026-04-17). Note this counts every `name = "..."` declaration in any `Cargo.toml` in the tree, which is a superset of `workspace.members` — `dora-examples` (the root package) is counted here but isn't a member. Every crate named in the plan is present in the tree. No phantom crate names.
 
 One nuance: plan §13 rename map specifies `shared-memory-server` → `dora-shared-memory-server` (for crates.io namespacing). The actual `libraries/shared-memory-server/Cargo.toml` still declares `name = "shared-memory-server"`. This is an **unexecuted rename action**, not a stale claim — the plan describes it as "will be renamed", which is still accurate.
 
@@ -171,16 +171,31 @@ No further independent review pass is warranted — the remaining items are poin
 Every command above is shell-one-liner. Full reproduction from a clean checkout of `main`:
 
 ```bash
-git checkout main
+# Fast-forward local main to origin/main so the subsequent file-exists
+# checks run against the same tree state the sweep recorded. A plain
+# `git fetch` alone doesn't move the working tree.
 git fetch origin main
+git checkout main
+git merge --ff-only origin/main
 
-# §1 File/path existence
-for path in apis/rust/compat/dora-node-api apis/rust/compat/dora-operator-api \
-            apis/rust/node/src/event_stream/data_conversion.rs \
-            apis/rust/node/src/node/mod.rs \
-            binaries/record-node binaries/replay-node binaries/ros2-bridge-node \
-            docs/audit-report-2026-03-21.md docs/migration-from-0.x.md \
-            examples/error-propagation examples/validated-pipeline; do
+# §1 File/path existence (16 paths — matches the 14-of-16 result in §1)
+for path in \
+    apis/rust/compat/dora-node-api \
+    apis/rust/compat/dora-operator-api \
+    apis/rust/node/src/event_stream/data_conversion.rs \
+    apis/rust/node/src/node/drop_stream.rs \
+    apis/rust/node/src/node/mod.rs \
+    binaries/record-node \
+    binaries/replay-node \
+    binaries/ros2-bridge-node \
+    docs/audit-report-2026-03-21.md \
+    docs/migration-from-0.x.md \
+    docs/ros2-bridge.md \
+    examples/error-propagation \
+    examples/log-sink-tcp \
+    examples/ros2-bridge/yaml-bridge-action-server/handler-node \
+    examples/ros2-bridge/yaml-bridge-action/goal-node \
+    examples/validated-pipeline; do
   [ -e "$path" ] && echo "EXISTS: $path" || echo "MISSING: $path"
 done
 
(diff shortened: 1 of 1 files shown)


heyong4725 (commented):
1. Medium: the sweep’s strongest “net-new finding” is wrong because it reads the wrong `serde_yaml` line from upstream `Cargo.toml`. In `docs/plan-consistency-sweep-2026-04-17.md:105`, the doc claims upstream’s actual value is `0.8.23` and concludes the plan is wrong on both `main` and `#286`. But upstream `Cargo.toml` has `serde_yaml = "0.9.33"` in `[workspace.dependencies]`; `0.8.23` only appears later in `[dev-dependencies]`. Since §3.2 is explicitly a workspace-level dependency table, this PR is flagging a false error and calling it “one of the most impactful findings” at `:115`.

2. Low: the upstream “workspace crate count = 39” finding is using the wrong metric. At `:88`, the method is `git/trees/... | count Cargo.toml files`, which counts every manifest in the repo, not `workspace.members`. The upstream root `Cargo.toml` currently has 35 `workspace.members` entries, so the document is comparing “workspace crates” on the fork against “all Cargo.toml files” upstream. That makes the table internally inconsistent and weakens the conclusion that the plan’s `~30` number is definitely wrong by the margin claimed.

heyong4725:
Both valid, both fixed in `2f3b05265`.

Sloppy method errors on my part — exactly what the sweep was supposed to catch.

**1. serde_yaml.** My grep took the last `^serde_yaml` line in upstream Cargo.toml, which is `0.8.23` at line 113 in `[dev-dependencies]`. Workspace-level value at line 89 is `0.9.33`. §3.2 is a workspace table, so `0.9.33` is correct and the plan row is a genuine 'tie'. Finding retracted. Sweep doc §6 updated to confirm plan is correct; §9 summary moved this item to a "Retracted" sub-section with the full cause-and-fix recorded so the error is reproducible-from-history.

**2. Workspace crate count.** I compared fork `workspace.members` (61) against upstream `git tree | grep Cargo.toml` (39). Apples-to-oranges. Real upstream `workspace.members` is 35 — which matches the plan's `~30` within approximation tolerance. Fork's 61 vs plan's `~45` remains a real undercount. Sweep doc §5 re-measured with the right method.

Also fixed §11 reproducing commands — they now use `awk '/^\[workspace.dependencies\]/,/^\[/'` scoping so the mistake isn't reproducible from the doc itself.

**Net findings 5 → 3 valid** (MSRV 1.88.0, fork workspace.members 61, stale GitHub metrics). 2 retracted.

Posting a correction on #286 too — round-5 feedback had the same two wrong claims and needs a retraction note so nothing gets incorporated from the wrong side.

heyong4725:
**/review round-2** on the corrected state. All three Codex findings valid, all three fixed in `f2d2663e4`.

1. **Crate count was 31; actual is 32.** §3. The command counts every `name = "..."` Cargo.toml declaration, not `workspace.members` — so it's a superset that includes the root `dora-examples` package. Updated the number and added a clarifying note.

2. **Reproduction loop missing 5 paths.** §11:177-183 listed 11 of the 16 paths §1 says were audited. Re-running the command could not reproduce the 14-of-16 result. Expanded to the full 16.

3. **Reproduction preamble didn't advance working tree.** §11 had `git checkout main; git fetch origin main` — a plain fetch doesn't move the tree, so a stale local main reproduces against different commits. Now `git fetch origin main; git checkout main; git merge --ff-only origin/main`.

**Also caught in self-review while fixing these:** the round-1 correction (2f3b05265) introduced an awk range `/^\[workspace.dependencies\]/,/^\[/` that exits immediately because the end pattern matches the start line. Swapped to explicit-flag form with a note explaining the gotcha.

**Pattern across all three rounds on this PR:** every one of my errors was a verification command that looked right but read slightly-wrong data — wrong awk range, wrong section, wrong count denominator, missing loop entry. The lesson for future sweep work: don't just check that the command parses, run it end-to-end and verify the output matches the claim before citing it.

**Verified after fixes:**
- §3 crate count 32: `grep -rh '^name = ' --include=Cargo.toml . | grep -oE '"dora-[^"]*"' | sort -u | wc -l` → 32 ✓
- §5 workspace.members 61 / 35: `sed -n '/^members = \[/,/^\]/p' Cargo.toml | grep -cE '^\s*"'` → 61; same on upstream → 35 ✓
- §6 serde_yaml 0.9.33 on both: `sed -n '/^\[workspace\.dependencies\]/,/^\[/p'` → 0.9.33 on both ✓
- §6 MSRV 1.85 / 1.88: `grep '^rust-version'` → fork 1.88.0, upstream 1.85.0 ✓

No further findings. Ready to merge.

PR #301, opened 2026-04-17
fix(qa): add explicit -g '*.rs' glob to unwrap-budget rg invocation

DGHX12345:
Adds `-g '*.rs'` as an explicit positive file-type glob alongside
`--type rust` in `scripts/qa/unwrap-budget.sh`.

**Why:** On some older ripgrep versions (pre-15), an `rg --files`
invocation with only negative globs and no explicit positive glob can
yield unexpected empty results in certain shell/PATH environments.
The extra glob is redundant with `--type rust` on rg ≥15 but harmless
— a defensive belt-and-suspenders guard.

**Scope:** Single-line change in `scripts/qa/unwrap-budget.sh`.
No behaviour change on rg ≥15.

## Test plan

- [x] Ran `bash scripts/qa/unwrap-budget.sh` locally — correct count
      returned (157, budget 185)
- [x] Diff is exactly 1 line

### DIFF
--- a/scripts/qa/unwrap-budget.sh
+++ b/scripts/qa/unwrap-budget.sh
@@ -53,6 +53,7 @@ count_unwraps() {
     total=$((total + n))
   done < <(
     rg --files --type rust \
+      -g '*.rs' \
       -g '!**/tests/**' \
       -g '!**/benches/**' \
       -g '!**/examples/**' \


heyong4725:
## /review on PR #301

**STATUS: DONE_WITH_CONCERNS — change is harmless but the rationale doesn't hold up, and the PR is low-value.**

### What the PR does
Adds `-g '*.rs'` to an existing `rg --files --type rust ...` invocation in `scripts/qa/unwrap-budget.sh`. +1 / -0.

### What's wrong with the justification

PR body says: *"On some older ripgrep versions (pre-15), an `rg --files` invocation with only negative globs and no explicit positive glob can yield unexpected empty results..."*

But the invocation **already has `--type rust`** as an explicit positive filter. So the premise ("only negative globs, no explicit positive glob") doesn't match the code. The belt-and-suspenders framing is for a scenario that isn't in the actual script.

### Empirical check on current ripgrep (15.1.0)

Ran both variants against the same tree (fresh main checkout):

```bash
rg --files --type rust libraries/ binaries/ apis/           → 252 files
rg --files --type rust -g '*.rs' libraries/ binaries/ apis/ → 252 files
```

Identical output. With the negative globs added, both produce the same 237-file pipeline into the awk counter. Final count: **157 (budget 185)** — same before and after this PR.

So on rg ≥15 the glob is a true no-op, as the PR body itself acknowledges. The question is whether the pre-15 concern is real.

### Is the pre-15 concern real?

I'm skeptical. Ripgrep's glob-handling semantics have been stable since well before 15.0. The scenario described ("only negative globs, no positive") doesn't match this code. If there's a specific rg version + env combination where `rg --files --type rust -g '!X'` returns empty, I'd want to see a repro or a linked ripgrep issue before trusting the mitigation.

### Side observation (not this PR's problem): budget is stale

The script reports `157 (budget 185)` and prints "Note: count reduced. Commit a smaller budget". The unwrap budget hasn't been ratcheted despite the count dropping by 28. Worth a follow-up commit bumping `.unwrap-budget` down — independently of this PR.

### Recommendation

Three options, in order of preference:

1. **Close** with thanks to the contributor. The PR's stated motivation doesn't hold up; the change is a pure no-op on current ripgrep and the "older versions" scenario doesn't match the actual invocation.

2. **Ask for evidence.** Request the contributor cite a specific rg version + shell combo where this matters, or a ripgrep issue link. If they can produce one, merge. If not, close.

3. **Merge as harmless defensive guard.** +1 line, no behaviour change on any version, adds a tiny bit of robustness for the theoretical edge case. Pro: cheap; engages external contributor. Con: normalizes speculative fixes without backing evidence.

I'd go with **1 or 2** over 3. Welcoming an external contributor is worth something, but merging a change on a misdescribed rationale trains bad habits. Better to close cleanly with a thank-you and point them to a higher-value issue from epic #287.

### Unrelated CI failure

Typos fails on `docs/plan-consistency-sweep-2026-04-17.md:221` (`mis-read`) — same unrelated main blocker from PR #276's review. I'll fix that typo in a separate 1-line PR so Typos stops blocking contributor PRs.

### Summary

| Item | Status |
|---|---|
| Stated rationale | Doesn't match the code (positive filter IS present) |
| Empirical behavior on rg 15.1 | No-op verified |
| Pre-15 concern | Speculative, no citation |
| Side finding: unwrap budget stale (185 → should be 157) | Not this PR's problem, flag as follow-up |
| Typos CI failure | Unrelated, my mess on main |

Closing or requesting-changes would be appropriate.

heyong4725:
Closing — see the [/review comment above](https://github.com/dora-rs/adora/pull/301#issuecomment-4268887416) for the full reasoning. TL;DR:

1. The PR body's premise ("only negative globs and no explicit positive glob") doesn't match the actual code — the `rg --files --type rust ...` invocation already has `--type rust` as an explicit positive filter.
2. Verified empirically on ripgrep 15.1: adding `-g '*.rs'` produces identical output (252 files both ways; 237-file pipeline into the awk counter either way; final count 157 unchanged).
3. The "pre-15 ripgrep edge case" justification is speculative — no ripgrep issue linked, no specific version/shell combo cited as a repro.

Appreciate the contribution and the careful test plan. For future PRs on this repo, the highest-value open work is tracked under epic #287 (dora 1.0 consolidation) — happy to point you at something more impactful if you'd like.

Unrelated note: the unwrap count has dropped from budget 185 → current 157. Opening a small follow-up to ratchet `.unwrap-budget` separately — that's a real, backed-by-evidence change worth landing.

PR #316, opened 2026-04-20
fix: use current_exe() instead of which("dora") to locate runtime binary

DGHX12345:
## Summary

The daemon was using `which::which("dora")` to resolve the binary path
for spawning runtime nodes. This resolves to whichever `dora` appears
first in `$PATH`, which may be a different version than the running
daemon — causing version skew between the daemon and the runtime it
spawns.

Changed to `std::env::current_exe()` so the daemon always invokes its
own binary with the `runtime` subcommand, guaranteeing version
consistency.

## Root cause

`binaries/daemon/src/spawn/spawner.rs` called `which::which("dora")`
while the same file already used `std::env::current_exe()` elsewhere
for other purposes. The inconsistency meant a system-installed `dora`
could be picked up instead of the locally-built one.

## Exposed by

Running `examples/c++-dataflow` with a locally-built binary while a
different version of `dora` was present in `$PATH` — runtime nodes
failed with "No such file or directory".

## Test results

- `examples/c++-dataflow`: runtime nodes spawn correctly,
  `internal counter: 1` on first tick ✅

### DIFF
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -297,8 +297,9 @@ impl Spawner {
                         ]);
                         Some(cmd)
                     } else {
-                        let mut cmd =
-                            Command::new(which::which("dora").wrap_err("failed to get dora path")?);
+                        let current_exe = std::env::current_exe()
+                            .wrap_err("failed to get current executable path")?;
+                        let mut cmd = Command::new(current_exe);
                         cmd = cmd.arg("runtime");
                         Some(cmd)
                     }


heyong4725:
@DGHX12345 thank you for this fix — it's a real bug and the fix is correct.

Unfortunately your PR landed on adora after this repository's 1.0 consolidation into `dora-rs/dora` (2026-04-17), and **the fix did not migrate**. The bug is still present on `dora-rs/dora` main at `binaries/daemon/src/spawn/spawner.rs` line 314.

I've opened a tracking issue on the canonical repo: **dora-rs/dora#NEW** (link in adora#316 comment after creation completes) crediting you as the original author. The cleanest path: re-open your PR against `dora-rs/dora` main with the same 3-line change. Apologies for the lost time — adora was supposed to be read-only for new work but the cutover communication was incomplete.

This repo is read-only post-consolidation; live work happens at [dora-rs/dora](https://github.com/dora-rs/dora).

heyong4725:
Follow-up: the dora-side tracking issue is **[dora-rs/dora#1797](https://github.com/dora-rs/dora/issues/1797)** (replacing the "#NEW" placeholder in the close comment above). @DGHX12345 the original PR diff is referenced verbatim in the issue body.

PR #318, opened 2026-04-21
fix: handle empty parent path when resolving dataflow working directory

DGHX12345:
## Summary

`Path::parent()` returns `Some("")` for bare filenames like
`"dataflow.yml"`, not `None`. The existing `.unwrap_or(Path::new("."))`
fallback only triggers on `None`, so `canonicalize("")` fails with
"No such file or directory" when the CLI tries to expand module
references.

Added `.filter(|p| !p.as_os_str().is_empty())` before the fallback
in all 6 affected call sites: `run`, `build`, `start`, `expand`,
`graph`, and `validate`.

## Root cause

Rust's `Path::parent()` on a relative path with no directory component
(e.g. `"dataflow.yml"`) returns `Some("")` rather than `None`. The
empty string cannot be canonicalized, causing module expansion to fail
at `libraries/core/src/descriptor/expand.rs:96`.

## Exposed by

Running `examples/module-dataflow` — `dora run dataflow.yml` from the
example directory triggers the empty base_dir path, failing with:

    failed to expand modules in dataflow descriptor
      failed to resolve base directory:
      No such file or directory (os error 2)

## Test results

- `examples/module-dataflow`: sender → doubler → filter → receiver
  pipeline runs correctly after fix ✅

### DIFF
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -145,6 +145,7 @@ pub fn build(
     }
     let working_dir = dataflow_path
         .parent()
+        .filter(|p| !p.as_os_str().is_empty())
         .unwrap_or_else(|| std::path::Path::new("."));
     let dataflow_descriptor = Descriptor::blocking_read(&dataflow_path)
         .wrap_err_with(|| {
--- a/binaries/cli/src/command/expand.rs
+++ b/binaries/cli/src/command/expand.rs
@@ -25,6 +25,7 @@ impl Executable for Expand {
         let working_dir = self
             .dataflow
             .parent()
+            .filter(|p| !p.as_os_str().is_empty())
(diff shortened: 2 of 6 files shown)


heyong4725:
@DGHX12345 thanks for the careful diagnosis on bare-filename handling!

The fix is **present on `dora-rs/dora` main** for 4 of the 6 files you patched: `expand.rs`, `validate.rs`, `start/mod.rs`, and `graph.rs` all have `.filter(|p| !p.as_os_str().is_empty())` exactly as in your PR.

`build/mod.rs` and `run.rs` were refactored on the dora side into a `BuildConfig` pattern that handles paths differently; the empty-parent issue is not present in the new code structure.

Closing as obsolete. If you spot residual cases on dora-rs/dora main after the refactor, please open a fresh issue there. Thanks!

